PR #4, opened 2023-08-01
Add ninja build steps

linsonittoop:


linsonittoop:
Discarding the PR as the email associated is not valid

PR #11, opened 2023-09-26
 Folder restructuring 

jvnaiduwin:


giorgiozoppi (commented):
Can we keep vcpkg support?

MelamudMichael:
> Can we keep vcpkg support?

Hi @giorgiozoppi 

For now we are going to support only redhat and yocto - so no need to vcpkg

MelamudMichael (approved):

jvnaiduwin:
> Can we keep vcpkg support?

As we have discussion with Michael, and he discussed with Steven we agreed upon 

"[Tuesday 1:29 PM] Michael Melamud

J Vignesh Naidu (C)  no need for vcpkg, we are supporting redhat and yocto at the moment 

[Tuesday 1:29 PM] Michael Melamud

this was the agreement with steven"

linsonittoop (approved):

stevenhartley (changes requested):
I pulled the project and built per the README and it fails to find RapidJSON even thought I installed it with vcpkg.  Can you please check what is wrong?  The build is complaining of a missing RapidJSONConfig.cmake or rapidjson-config.cmake file

uprotocol/tools/tools.h
stevenhartley, line 1: Would this not be provided by some other external library?

uprotocol/proto/CMakeLists.txt
stevenhartley, line 1: We should not include the cloudevent proto here, it should be pulled as an external dependency for the cloudevent builders.

stevenhartley (changes requested):
additional comments

stevenhartley (approved):
Change is now good and we can merge to the 1.5 branch

PR #36, opened 2024-01-17
aligning

MelamudMichael:


stevenhartley (changes requested):
@xz62px please pull a release tag and not main from uprotocol-core-api, that way you can coordinate the change

stevenhartley:
> > @xz62px please pull a release tag and not main from uprotocol-core-api, that way you can coordinate the change
> 
> @stevenhartley Can you give me an example what do i need to do ?

Please take a look at the Java, Kotlin, and Python projects, they are doing a git clone of a specific tag of core-api.

stevenhartley:
Per @MelamudMichael closing PR, will be merged as part of another PR (@adienzel uri 1.5 changes)

Issue #2, opened 2023-06-16
C++ SDK: Initial Commit


stevenhartley:
Initial drop that matches the functionality of the Java SDK

stevenhartley:
Commit done

PR #43, opened 2024-02-05
Feature/conan support #2

MelamudMichael:


stevenhartley (approved):
Will fix README in a follow on commit

PR #45, opened 2024-02-06
Urivalidator Implementation.

jvnaiduwin:
This will validate Uri on basis on the Uri passed, whether its valid or not and return boolean or UStatus.

Note : 1. 8 test case differ in result when comparing to Java.
           2. LongUriSerializer deserialize API doesn't provide result as per Java sdk.


stevenhartley:
Pr has gone stale, closing

PR #46, opened 2024-02-06
[cmake] Keep unit tests as optional

uilianries:
When packaging Up-cpp (e.g. Conan), unit tests are not needed, so the usual flow is skipping it.

This PR adds the option `UP_CPP_BUILD_TESTING`, and supports the native CMake option `BUILD_TESTING` (https://cmake.org/cmake/help/latest/command/enable_testing.html)

It will build tests by default, but will give the option to skip it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,10 @@ else()
 						            spdlog::spdlog)
 endif()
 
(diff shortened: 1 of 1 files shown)


stevenhartley:
Thank you @uilianries !

PR #47, opened 2024-02-06
[cmake] Add option for up-core-api root directory

uilianries:
When building up-cpp, the up-core-api should be installed too, but as same root level.

This PR gives the option to use up-core-api from another path. It's useful in case up-core-api is install via a package manager and is not copied to the same place.

Plus, when up-core-api is not found, the project should stop. So, the CMake FATAL_ERROR stops the execution as an error message.


stevenhartley:
Thank you @uilianries !

PR #61, opened 2024-03-05
Uprotocol 1.5

MelamudMichael:
The PR is adding support for uattributes from proto , and adds a new uattributes builder class


include/up-cpp/transport/builder/UAttributesBuilder.h
stevenhartley, line 1: What we need is uProtocol based uAttributes builder meaning we build publish, notification, request, and response headers, otherwise you leave the complexity of building these types (and ensuring the right information is populated) up to the app developer who will more often than not miss something that is mandatory.

include/up-cpp/transport/datamodel/UAttributesBuilder.h
stevenhartley, line 1: why are there 2 attribute builders?

the date is wrong in the file, need to be 2024

stevenhartley (changes requested):
Please see comments in the review

PR #64, opened 2024-03-10
Adding support for full build 

MelamudMichael:
1. Adding support for full build
2. Enabling uuid tests

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -16,7 +16,7 @@ class UpCpp(ConanFile):
     options = {"shared": [True, False], "fPIC": [True, False]}
     conan_version = None
     generators = "CMakeDeps", "PkgConfigDeps", "VirtualRunEnv", "VirtualBuildEnv"
-    version = "0.1.5.1-dev"
+    version = "0.1.1-dev"
     exports_sources = "CMakeLists.txt", "up-core-api/*", "include/*" ,"src/*" , "test/*", "cmake/*"
 
     options = {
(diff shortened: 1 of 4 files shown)


conanfile.py
```diff
@@ -16,7 +16,7 @@ class UpCpp(ConanFile):
     options = {"shared": [True, False], "fPIC": [True, False]}
     conan_version = None
     generators = "CMakeDeps", "PkgConfigDeps", "VirtualRunEnv", "VirtualBuildEnv"
-    version = "0.1"
+    version = "0.1.5.1-dev"
```
PLeVasseur, line 19: Should this be in sync with the current `up-core-api` release supported? Is `up-cpp` at `1.5.1` or a later tag?

billpittman (approved):
Worked for me just fine.

conanfile.py
MelamudMichael: @stevenhartley  what version do we want to put here?

stevenhartley: 0.1.x-dev for now. Once we pass RC3 we can consider naming it 1.x.x.

MelamudMichael: so thats fine? i set it to 0.1.5.1-dev  because the current tag is [v0.1.5.0-dev](https://github.com/eclipse-uprotocol/up-cpp/releases/tag/v0.1.5.0-dev) , so the next tag will be 0.1.5.1 

If you are ok with the change please merge the PR

stevenhartley: That is my fault, it should not have been x.x.x.x but follow semver, that is why I say we should be v0.1.x-dev

MelamudMichael: Done !

PR #68, opened 2024-03-17
Feature/rpc alignment

MelamudMichael:
1. Aligning protobufs version to 1.5.7
2. Sending attachment headers as a protobuf object
3. Update RPC API definitions


include/up-cpp/rpc/RpcClient.h
```diff
@@ -19,42 +19,68 @@
...
-        * @param attributes Metadata for the method invocation (i.e. priority, timeout, etc.)
-        * @return Returns the CompletableFuture with the result or exception.
-        */
-        virtual std::future<UPayload> invokeMethod(const UUri &topic, 
-                                                   const UPayload &payload, 
-                                                   const UAttributes &attributes) = 0;
-        
-        virtual ~RpcClient() {} 
-};
+    struct RpcResponse {
```
PLeVasseur, line 36: I saw that C++17 is being used, so how about using a [`std::variant`](https://en.cppreference.com/w/cpp/utility/variant) which could express that the `RpcResponse` is one or the other?

For example:
```cpp
using RpcResponse = std::variant<uprotocol::utransport::UMessage, uprotocol::v1::UStatus>;
```

Client in uE code could then do one of two things:
1. Use [`get_if()`](https://en.cppreference.com/w/cpp/utility/variant/get_if) in order to check for the one that's been filled
2. Use the [visitor pattern](https://en.cppreference.com/w/cpp/utility/variant/visit) (should probably recommend this route in docs / examples -- it's safer)

If C++23 were on the table (I don't think so...?), then it'd be possible to use C++23's [`std::expected`](https://en.cppreference.com/w/cpp/utility/expected) directly, which better semantically represents what we want of a 'typical' and 'non-typical' value. (side note: was not aware that had been implemented for C++23! so cool. convergent evolution :slightly_smiling_face:)

Changelog
```diff
@@ -1,9 +1,9 @@
 ## Changelog
 
-- June 04 2023 17:52:23  Giorgio Zoppi <giorgio.zoppi@iotdbe.com>
+- June 04 2024 17:52:23  Giorgio Zoppi <giorgio.zoppi@iotdbe.com>
```
PLeVasseur, line 3: I'm guessing this line and the line below got hit by a `2023` => `2024` find and replace all

include/up-cpp/transport/UTransport.h
```diff
@@ -51,9 +48,9 @@ namespace uprotocol::utransport {
                * @return Returns OKSTATUS if the payload has been successfully sent (ACK'ed), otherwise it
                * returns FAILSTATUS with the appropriate failure.
                */
-               virtual UStatus send(const UUri &uri, 
-                                    const UPayload &payload,
-                                    const UAttributes &attributes) = 0;
+               virtual uprotocol::v1::UStatus send(const uprotocol::v1::UUri &uri,
```
PLeVasseur, line 51: According to [the spec](https://github.com/eclipse-uprotocol/up-spec/tree/main/up-l1#21-send), the signature should be:
```
OUT UStatus send(UMessage&)
```

billpittman (changes requested):

include/up-cpp/rpc/RpcClient.h
MelamudMichael: Not everyone will be using C++17 , for example CHM

Changelog
MelamudMichael: Or giorgio is working from the future

include/up-cpp/transport/UTransport.h
MelamudMichael: Yes, David is working on this

include/up-cpp/transport/datamodel/UPayload.h
```diff
@@ -39,124 +39,118 @@ namespace uprotocol::utransport {
...
-                    data_ = nullptr;
-                    refPtr_ = nullptr;
-                    dataSize_ = 0;
+                if (type == UPayloadType::REFERENCE) {
+                    dataPtr_ = std::shared_ptr<const uint8_t[]>(ptr, [](const uint8_t*){});
+                } else {
+                    dataPtr_ = std::shared_ptr<const uint8_t[]>(new uint8_t[dataSize_], [](const uint8_t* p){ delete[] p; });
+                    if (nullptr != ptr) {
+                        std::memcpy(const_cast<uint8_t*>(dataPtr_.get()), ptr, dataSize_);
+                    }
```
gregmedd, line 71: Is this an interface that would be used by consumers of this library? I'm a little concerned about this interface/implementation (bare pointers with no ownership semantics, not using containers with included size tracking, allocating memory for a shared_ptr using new instead of make_shared, etc). If it's purely internal to the library and not meant to be used by consumers, then it's probably fine for now. Otherwise if developers will be writing code to this, we should probably work on a cleaner interface.

include/up-cpp/transport/datamodel/UPayload.h
```diff
@@ -39,124 +39,118 @@ namespace uprotocol::utransport {
...
+                     const UPayloadType &type) : dataSize_(size), type_(type) {
 
-                if ((nullptr == data) || (0 == dataSize)) {
-                    data_ = nullptr;
-                    refPtr_ = nullptr;
-                    dataSize_ = 0;
+                if (type == UPayloadType::REFERENCE) {
+                    dataPtr_ = std::shared_ptr<const uint8_t[]>(ptr, [](const uint8_t*){});
+                } else {
+                    dataPtr_ = std::shared_ptr<const uint8_t[]>(new uint8_t[dataSize_], [](const uint8_t* p){ delete[] p; });
```
gregmedd, line 68: Consider using `std::make_shared<std::vector>(dataSize_)` instead

(NB: C++20 would allow the use of `std::make_shared<const uint8_t[]>(dataSize_)`)

include/up-cpp/rpc/RpcClient.h
```diff
@@ -19,42 +19,68 @@
...
+            * @param methodUri The method URI to be invoked, ex (long form): /example.hello_world/1/rpc.SayHello.
+            * @param requestPayload The request message to be sent to the server.
+            * @param options RPC method invocation call options, see {@link CallOptions}
+            * @param callback that will be called once the future is complete
+            * @return UStatus
+            */
+            virtual uprotocol::v1::UStatus invokeMethod(const uprotocol::v1::UUri &topic,
+                                                        const uprotocol::utransport::UPayload &payload,
+                                                        const uprotocol::v1::CallOptions &options,
+                                                        const uprotocol::utransport::UListener &callback) = 0;
```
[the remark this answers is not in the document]
MelamudMichael: by mistake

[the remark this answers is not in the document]
MelamudMichael: @gregmedd  can you open an issue on this to discuss?

include/up-cpp/transport/datamodel/UPayload.h
MelamudMichael: UPayload used as a container , basically you set the buffer you want to send to UPayload.

Could you please provide a code sample for the changes ?

cmake/up-core-api-protos.cmake
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2023 General Motors GTO LLC
+# Copyright (c) 2024 General Motors GTO LLC
```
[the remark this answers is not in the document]
gregmedd: > Should we be updating copyright dates in files that haven't otherwise been changed? Copyrights are valid for _decades_ in most places, so doing a global find and replace in a repo just results in excess file churn.

I don't know why this was marked as resolved. Copyright date updates in files that were not otherwise modified should be reverted to minimize the number of unnecessary changes in this PR.

include/up-cpp/rpc/RpcClient.h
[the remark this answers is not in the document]
gregmedd: @MelamudMichael - opened as #78. If we go the route of supporting both, `std::function` can be added later without breaking existing client code.

include/up-cpp/transport/datamodel/UPayload.h
gregmedd: [Here's an example](https://github.com/MelamudMichael/up-cpp/compare/feature/rpc_alignment...gregmedd:up-cpp:example/UPayload-without-bare-pointers)

include/up-cpp/transport/builder/UAttributesBuilder.h
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ * 
+ * SPDX-FileType: SOURCE
+ * SPDX-FileCopyrightText: 2023 General Motors GTO LLC
```
stevenhartley, line 22: 2024

include/up-cpp/transport/builder/UAttributesBuilder.h
```diff
@@ -0,0 +1,147 @@
...
+
+#include <string>
+#include <spdlog/spdlog.h>
+#include <up-cpp/uuid/factory/Uuidv8Factory.h>
+#include <up-core-api/uattributes.pb.h>
+#include <up-core-api/uuid.pb.h>
+
+namespace uprotocol::utransport {
+
+class UAttributesBuilder {
```
stevenhartley, line 37: You need to document the interface as you expect developers to use this API

include/up-cpp/transport/builder/UAttributesBuilder.h
```diff
@@ -0,0 +1,147 @@
...
+                           uprotocol::v1::UPriority priority) {
+
+            attributes_.Clear();
+            attributes_.mutable_id()->CopyFrom(id);
+            attributes_.mutable_source()->CopyFrom(source);
+            attributes_.set_type(type);
+            attributes_.set_priority(priority);
+        }
+
+        UAttributesBuilder& setToken(const std::string& token) {
```
stevenhartley, line 58: Methods need documentation, CoPilot can autogenerate them for you, please use it.

include/up-cpp/transport/datamodel/UPayload.h
```diff
@@ -39,124 +39,119 @@ namespace uprotocol::utransport {
         UNDEFINED  /* invalid */
     };
 
+    // The Serialization format for the data stored in UPayload.
+    enum UPayloadFormat {
```
stevenhartley, line 43: Please use the protobuf defined UPayloadFormat and NOT this defined enum so there are no issues with values being out of sync.

test/uri/builder/UAuthorityTest.cpp
```diff
@@ -1,6 +1,6 @@
 /*
  
- * Copyright (c) 2024 General Motors GTO LLC
+ * Copyright (c) 2023 General Motors GTO LLC
```
stevenhartley, line 3: Please back out this change if you did not change the file

test/uri/builder/UEntityTest.cpp
```diff
@@ -1,6 +1,6 @@
 /*
  
- * Copyright (c) 2024 General Motors GTO LLC
+ * Copyright (c) 2023 General Motors GTO LLC
```
stevenhartley, line 3: backout this change

test/uri/builder/UResourceTest.cpp
```diff
@@ -1,6 +1,6 @@
 /*
  
- * Copyright (c) 2024 General Motors GTO LLC
+ * Copyright (c) 2023 General Motors GTO LLC
```
stevenhartley, line 3: backout this change

test/uri/builder/UUriTest.cpp
```diff
@@ -1,6 +1,6 @@
 /*
  
- * Copyright (c) 2024 General Motors GTO LLC
+ * Copyright (c) 2023 General Motors GTO LLC
```
stevenhartley, line 3: backout this change

billpittman (approved):

test/utransport/uattributes_test.cpp
stevenhartley, line 1: This test only creates one UAttributes object and doesn't test all the positive and negative flows. You need 100% code coverage for what you're adding (UAttributeBuilder).

test/utransport/upayload_test.cpp
stevenhartley, line 1: Same comment with the other test, there is no coverage

stevenhartley (changes requested):

gregmedd (approved):
If we file an issue to address the documentation and test of `UAttributesBuilder.h` in a follow-up PR, and likewise file an issue to discuss my interface recommendations on `UPayload.h`, then this PR looks ready to move forward.

include/up-cpp/transport/builder/UAttributesBuilder.h
debruce: Should this be 2024 or 2023-2024?

stevenhartley (approved):
This Pr is blocking a lot of other PRs from being merged so we will merge it and fix all the open issues in separate smaller PRs.

PR #74, opened 2024-03-27
Feature/1 5 7 alignment

MelamudMichael:
Aligning protobufs version to 1.5.7
Sending attachment headers as a protobuf object
Update RPC API definitions


gregmedd:
Is this different than the existing PR (#68)? It looks the same other than by branch name (and maybe one or two commits). Is the intent to complete #68 first, then follow-up with this PR?

Part of the reason I ask is that there are several existing comments on #68 that do not appear to have been discussed or addressed yet.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
- # Copyright (c) 2023 General Motors GTO LLC
+ # Copyright (c) 2024 General Motors GTO LLC
```
gregmedd, line 1: Why are copyright dates being updated in files that have not otherwise changed? By my count, _at least_ 34 of the 62 files changed in this PR _only_ update the copyright year. Unless there is a good reason to do this, all 34 of these files should all be reverted.

Changelog
```diff
@@ -1,9 +1,9 @@
 ## Changelog
 
-- June 04 2023 17:52:23  Giorgio Zoppi <giorgio.zoppi@iotdbe.com>
+- June 04 2024 17:52:23  Giorgio Zoppi <giorgio.zoppi@iotdbe.com>
 	* Restructuring old code and deleting unused code.
     * Adding coverage and ctest support
-- June 10 2023 17:52:23  Giorgio Zoppi <giorgio.zoppi@iotdbe.com>
+- June 10 2024 17:52:23  Giorgio Zoppi <giorgio.zoppi@iotdbe.com>
```
gregmedd, line 6: This should not have been changed. Please revert.

src/uri/IpAddress.cpp
gregmedd, line 1: This change should be merged through the separate, existing PR (#73) and not mashed into this one. Revert.

gregmedd (changes requested):
I'm going to hold off on further review of this change until we discuss the necessity of replacing #68 with this PR. IMO, #68 should be reopened so the review comments there can be addressed and the bulk changes merged first. I'm not sure I see the benefit of starting the review over from scratch here. No significant changes have been made here as far (as I can tell) other than pulling in additional changes that were already represented in separate PRs.

PR #75, opened 2024-03-28
Add support for the Zenoh Client's usage of the micro URI

gregmedd:
[Per spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/up-l1/zenoh.adoc#22-uri-mapping), the Zenoh client uses the micro URI serialization for UUri. However, it has support for one extra `**` UUri mode that requires the ability to serialize _only_ the UAuthority.

This change splits out the UAuthority serialization into its own static method, which can be used both when serializing full UUris and when generating Zenoh keys for the `**` mode.

~~Note: This PR is dependent on #73 and cannot be merged until that PR is complete.~~

### DIFF
--- a/include/up-cpp/uri/serializer/MicroUriSerializer.h
+++ b/include/up-cpp/uri/serializer/MicroUriSerializer.h
@@ -55,6 +55,13 @@ public:
     [[nodiscard]] static auto serialize(const uprotocol::v1::UUri& u_uri) -> std::vector<uint8_t>;
 
     /**
+     * Serialize a UAuthority into a vector<uint8_t> following the Micro-URI specifications.
+     * @param u_auth The UAuthority data object.
+     * @return Returns a vector<uint8_t> representing the serialized UAuthority.
+     */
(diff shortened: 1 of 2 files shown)


gregmedd:
This PR can be reviewed now. However, only the last commit ([8d1fac3](https://github.com/eclipse-uprotocol/up-cpp/pull/75/commits/8d1fac3311ce23f7d270309584c24780dd2efb36)) should be reviewed. All commits before that are from #73. Once #73 is merged, I'll take this out of draft state so it's clear that it is ready to go.

debruce (approved):
I think this looks okay, and surely represents a large amount of work. However, I think it points to wider design issues, especially around the use of static methods instead of constructors, error propagation, absence of hashing methods for objects likely to be used as keys, etc.

gregmedd:
> ... surely represents a large amount of work.

> This PR can be reviewed now. However, only the last commit ([8d1fac3](https://github.com/eclipse-uprotocol/up-cpp/pull/75/commits/8d1fac3311ce23f7d270309584c24780dd2efb36)) should be reviewed. All commits before that are from #73. Once #73 is merged, I'll take this out of draft state so it's clear that it is ready to go.

stevenhartley (approved):

PR #96, opened 2024-04-18
Clarifying conan responsibility of source, install, build, create

PLeVasseur:
# What

  * conanfile.py contains the Conan recipes to build & package
  * updated conanfile.py to add source() method which clones the .proto holding repository
  * updated up-core-api-protos.cmake to appropriately refer to location of .proto definitions set in conanfile.py, making it future proof
  * updated GitHub workflow for new build steps
  * updated GitHub workflow to separate test & package flows

# Why

We want to support local builds & packaging.

# To Dos

- [ ] Update README.md to discuss how to make releases to CCI
- [ ] Confirm that the CCI package can be accepted into CCI

# Next Steps

We also want to separately create an entry into [Conan Center Index](https://github.com/conan-io/conan-center-index) to support pulling `up-cpp` packages directly for users.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,8 @@ on:
   workflow_dispatch:
   
 jobs:
-  build:
-    name: Build
+  test:
+    name: Test
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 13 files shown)


stevenhartley (changes requested):
@PLeVasseur you need to fix/update the ci workflows with the changes to the build instructions, otherwise the workflows will fail and we cannot merge the change

PLeVasseur:
Pause on this while I update the GitHub workflow a little more

.github/workflows/ci.yml
```diff
@@ -37,21 +35,48 @@ jobs:
...
           ctest
 
       - name: Create up-cpp Conan package
         shell: bash
         run: |
-          conan create . --build=missing
+          cd conan-recipes
+          conan create .. --build=missing
+
+  package:
```
PLeVasseur, line 52: @stevenhartley, @gregmedd -- FYI I updated to separate out the test vs package flows to distinguish between them and their conanfile.py

.github/workflows/ci.yml
```diff
@@ -62,7 +87,7 @@ jobs:
   ci:
     name: CI status checks
     runs-on: ubuntu-latest
-    needs: build
+    needs: [test, package]
```
PLeVasseur, line 90: Depending on the new package flow as well

PLeVasseur:
> Pause on this while I update the GitHub workflow a little more

OK -- should be good for review :slightly_smiling_face:

conanfile_package.py
stevenhartley, line 1: why rename and not in separate file (out of curiosity)?

PLeVasseur: Good question!

I tried to put it in `conan-recipes`, but seems like Conan forbids you from using ".." to hop out of the folder for security reasons.

So this was the next best alternative I found.

PLeVasseur:
Putting this in draft status while I try to work out the kinks a little more.

Next goal:
* version which works for local build & package
* version which can be put into CCI

PLeVasseur:
@stevenhartley -- broke out into a conanfile.py we can use for local builds & packaging and another one under the ./conan-recipes/cci that we can use for uploading to CCI.

Both work and have workflows added for them.

Need to open a PR over to CCI and see what else may be wrong with the CCI one.

PLeVasseur:
> Putting this in draft status while I try to work out the kinks a little more.
> 
> Next goal:
> 
> * version which works for local build & package
> * version which can be put into CCI

Well Conan the package manager has the perfect tool for the use case I'm considering of sharing packaging instructions between the CCI recipe and local build recipe: python requires

However, recipes hosted on CCI are banned from using it :upside_down_face: 

https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#are-python-requires-allowed-in-the-conan-center-index

Given that path is a dead-end I tried to inherit from the CCI recipe, but that's looking like it's tricky as well, because the relative paths between the CCI recipe and the local one are not maintained when copied into the Conan cache (`~/.conan2`).

Based on that, I'm considering the current form of this PR to be ~ reasonable enough ~ where-in we have 2 x recipes, one for CCI packaging and one for local build & package without doing anything fancy to share code.

WDYT @stevenhartley, @gregmedd ?

PLeVasseur:
FYI all -- I submitted the package to CCI, waiting to hear feedback and reflect that into the one we'll maintain in the repo.

gregmedd:
Closing stale PRs. We are tracking conan recipes on the project board.

PR #97, opened 2024-04-19
SetID: ID is allowed to be 0 in these cases

billpittman:
Note: This should be merged before #34

### DIFF
--- a/include/up-cpp/uri/builder/BuildUResource.h
+++ b/include/up-cpp/uri/builder/BuildUResource.h
@@ -95,11 +95,7 @@ namespace uprotocol::uri {
          * @return
          */
         auto setID(const uint32_t &id) -> BuildUResource & {
-            if (0 == id) {
-                spdlog::error("UResource id cannot be 0");
-            } else {
-                resource_.set_id(id);
-            }
+            resource_.set_id(id);
             return *this;
         }
         /**
@@ -140,12 +136,8 @@ namespace uprotocol::uri {
          * @return 
          */
         auto setRpcRequest(const uint32_t id) -> BuildUResource & {
-            if (0 == id) {
-                spdlog::error("UResource id cannot be 0");
-            } else {
-                setName("rpc");
-                setID(id);
-            }
+            setName("rpc");
+            setID(id);
             return *this;
         }
         /**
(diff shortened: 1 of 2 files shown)


include/up-cpp/uri/builder/BuildUResource.h
```diff
@@ -140,12 +136,8 @@ namespace uprotocol::uri {
...
          */
         auto setRpcRequest(const uint32_t id) -> BuildUResource & {
-            if (0 == id) {
-                spdlog::error("UResource id cannot be 0");
-            } else {
-                setName("rpc");
-                setID(id);
-            }
+            setName("rpc");
+            setID(id);
```
stevenhartley, line 140: If the id is 0, you need to call setRpcResponse() so that the name and instance is also populated correctly (incase someone called this API in lieu of setRpcResponse()

gregmedd: If I'm understanding this, the ask would be to undo the changes here in `setRpcRequest` since the request doesn't use ID 0. The _response_ uses ID 0 though, so the changes to `setID()` are needed.

@stevenhartley - is that what you are asking for?

gregmedd (changes requested):
@billpittman - The [unit tests failed](https://github.com/eclipse-uprotocol/up-cpp/pull/97/checks#step:8:14). Looks like the tests for this file were checking that ID 0 was disallowed. Can you update the test?

include/up-cpp/uri/builder/BuildUResource.h
billpittman: Wait, so if someone calls setRpcRequest with an ID of 0, then we "assume" they meant setRpcResponse, and call it for them ?  Do I have that right?

gregmedd: My interpretation of the ask is for the only change in this file to be the one on line 98. This change here in `setRpcRequest()` is not needed because a _request_ cannot have an ID of 0.

gregmedd (approved):
LGTM. Should be good to go, pending resolution of @stevenhartley's comment.

stevenhartley (approved):
Unfortunately there are many errors in the uProtocol logic in these test cases not to mention numerous errors in the builders for building topics requests, repsonses. Another issue will be files to re-write large portions of the cpp library. Only approving so we can have something that compiles.

stevenhartley:
Pains me to approve this change mainly because there is so much else that needs to be re-done in the builder and all the test cases :-(

Issue #102, opened 2024-05-08
MicroUriSerializer check has incorrect logic


ALPayne18:
When a Micro URI is validated, it assumes that the entity.id() must be greater than 0.  
https://github.com/eclipse-uprotocol/up-cpp/blob/494b3313b55aaefe5b8b0376dd761b75fa96a78a/include/up-cpp/uri/serializer/MicroUriSerializer.h#L193C1-L195C6

However, this is not the case. The core uSubscription uService has an id == 0: 

```
// Subscription Service Interface definition
service uSubscription {
  option (name) = "core.usubscription"; // Service name
  option (version_major) = 3;
  option (version_minor) = 0;
  option (id) = 0;

```
This bug causes any URI for the uSubscription service to fail to serialize.

PLeVasseur:
Hi @ALPayne18 --

This _does_ look to be a legit bug.

I can find nowhere in the [Micro UUri](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc#43-micro-uris) spec mention of restricting the UEntity id to be non-zero.

Further review of `up-rust` and `up-java` also shows no such restriction in-place.

@gregmedd -- how'd we like to handle this?

gregmedd:
The spec explicitly calls out an ID of 0 as being for uSubscription, so I agree it needs to be fixed. Is @debruce or @billpittman available to grab this issue and put out a PR with a fix?

gregmedd:
This has been resolved with #104. Do we need to tag a release, or do we have other bugfixes?

billpittman:
FWIW, this has been merged and theoretically fixed, but I can't assign it to myself, nor close it.  However it should be at least closed.

PR #104, opened 2024-05-10
MicroUriSerializer: Allow Id to be zero

billpittman:
Previously zero was considered an invalid Id and was causing problems with uSubscription calls since the spec specifically calls out an ID of 0 being for uSubscription.

Addresses issue: https://github.com/eclipse-uprotocol/up-cpp/issues/102


PLeVasseur (changes requested):
Hey @billpittman -- thanks for taking this on.

Looks to me like something is failing in CI. Could you take a look?

billpittman:
Looks like gtest 1.4 is required but somehow not part of the conancenter or download or something.  I'm guessing some sort of connectivity or conan issue.

I also noticed that the eclipse/eca isn't passing, but it totally SHOULD be, I have an active agreement.

gregmedd (approved):
LGTM

PR #105, opened 2024-05-17
New API - Post review deployment

gregmedd:
API review was completed in #99. However, that review was structured for visibility and cannot be merged directly.

This PR contains the approved API along with a buildable skeleton of the up-cpp library + tests. This aims to merge with main.

(Note: in draft state while verifying CI works still and finalizing a couple small items.)


conanfile.txt
```diff
@@ -0,0 +1,15 @@
+[requires]
+up-core-api/1.5.8
```
PLeVasseur, line 2: This is probably a good thing, as from what I've been reading, CCI tends to want to see only other Conan packages used as dependencies, not pulling in git submodules or doing git clones.

include/up-cpp/README.md
```diff
@@ -0,0 +1,33 @@
...
+* `validator/` - Functions for inspecting and validating the content of uProtocol objects
+
+## L1: Transport
+
+The abstract API in `transport/` is for sending and receiving uProtocol
+messages over a transport protocol, such as Zenoh or SOME/IP.
+
+## L2: Communication
+
+uEntities building on uProtocol will use the APIs in `communication/`.  These
```
PLeVasseur, line 25:
```suggestion
uEntities building on uProtocol will typically use the APIs in `communication/`.  These
```

?

include/up-cpp/communication/NotificationSink.h
```diff
@@ -0,0 +1,80 @@
...
+
+/// @brief Interface for uEntities to receive notifications.
+///
+/// Like all L2 client APIs, the NotificationSink is a wrapper on top of the L1
+/// UTransport API; in this instance, it provides for the notification
+/// receiving half of the notification model.
+struct NotificationSink {
+	using ListenCallback = transport::UTransport::ListenCallback;
+
+	using StatusOrSink =
```
PLeVasseur, line 35:
```suggestion
	using SinkOrStatus =
```

?

Matches the ordering of the types the `Expected` is templated over.

include/up-cpp/communication/NotificationSink.h
```diff
@@ -0,0 +1,80 @@
...
+	/// @param callback Called when a notification is received.
+	/// @param source_filter (Optional) URI to compare against notification
+	///                      sources. Only notifications that match will be
+	///                      forwarded tot he callback.
+	///
+	/// @returns
+	///    * unique_ptr to a NotificationSink if the callback was connected
+	///      successfully.
+	///    * UStatus containing an error state otherwise.
+	static StatusOrSink create(const v1::UUri& sink, ListenCallback&& callback,
```
PLeVasseur, line 55: Did this also need to take a `std::shared_ptr<transport::UTransport> transport`? I'm unclear otherwise on how to create a `NotificationSink`, given that the constructor is `protected`.

include/up-cpp/communication/NotificationSource.h
```diff
@@ -0,0 +1,75 @@
...
+/// sending half of the notification model.
+struct NotificationSource {
+	/// @brief Constructs a notification source connected to a given transport.
+	///
+	/// @post An internal UMessageBuilder will be configured based on the
+	///       provided attributes.
+	///
+	/// @param transport Transport to publish messages on.
+	/// @param sink URI of this uE. The authority and entity will be replaced
+	///             automatically with those found in the transport's default.
```
PLeVasseur, line 41: By default, do you mean [this](https://github.com/eclipse-uprotocol/up-cpp/pull/105/files#diff-a36088a3d7aefa487d703f85584902256a8d27e3daec98f6cdaa6edced6d44a6R52)?

I'm unsure of a nicer way to tie these concepts together.

include/up-cpp/communication/NotificationSource.h
```diff
@@ -0,0 +1,75 @@
...
+	/// @param sink URI of the uE notifications will be sent to.
+	/// @param payload_format (Optional) If sending a payload, this sets the
+	///                       format that will be expected when notify() is
+	///                       called. Empty response payloads can only be sent
+	///                       if this was not set.
+	/// @param priority All sent notifications will be assigned this priority.
+	/// @param ttl How long messages will be valid from the time notify() is
+	///            called.
+	NotificationSource(std::shared_ptr<transport::UTransport> transport,
+	                   const v1::UUri& source, const v1::UUri& sink,
```
PLeVasseur, line 51: I think I asked this on the other PR, but maybe forgot about it -- so does this mean for the source UUri, the only info taken from it and used is the ue_version_major and the resource_id?

It seems to be the case, but I'm just now catching up with the UUri changes.

gregmedd: Yes, I believe that is the intent. Unless there is a usecase for overriding the defaults provided by the UTransport

include/up-cpp/datamodel/builder/Payload.h
```diff
@@ -0,0 +1,159 @@
...
+#include <uprotocol/v1/uattributes.pb.h>
+
+#include <cstdint>
+#include <string>
+#include <tuple>
+#include <vector>
+
+namespace uprotocol::datamodel::builder {
+
+/// @brief Interface for preparing payloads for inclusing in a UMessage
```
PLeVasseur, line 24:
```suggestion
/// @brief Interface for preparing payloads for inclusion in a UMessage
```

?

include/up-cpp/communication/NotificationSink.h
gregmedd: Yes. Yes it does.

include/up-cpp/communication/RpcClient.h
```diff
@@ -0,0 +1,101 @@
...
+	///                       format that will be expected when invokeMethod()
+	///                       is called. Empty request payloads can only be
+	///                       sent if this was not set.
+	/// @param permission_level Permission level of this client
+	/// @param token TAP token for accessing restricted services
+	///
+	/// For guidance on the permeission_level and token parameters, see:
+	/// https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/permissions.adoc
+	explicit RpcClient(std::shared_ptr<transport::UTransport> transport,
+	                   v1::UUri&& method, v1::UPriority priority,
```
PLeVasseur, line 51: What do you think about for the constructors for the uP-L2 objects taking a configuration parameter, e.g. `RpcClientConfig`?

I see the parameter list growing a bit on this one in particular growing a bit long.

Not clear if this is considered good form or not, though.

include/up-cpp/communication/RpcClient.h
```diff
@@ -0,0 +1,101 @@
...
+	                   std::optional<uint32_t> permission_level = {},
+	                   std::optional<std::string> token = {});
+
+	/// @brief as found in v1::UAttributes
+	using Commstatus = uint32_t;
+
+	/// @brief Contains either a UMessage (when successful) or a UStatus /
+	/// Commstatus when an error occurred.
+	using StatusOrMessage =
+	    utils::Expected<v1::UMessage, std::variant<v1::UStatus, Commstatus>>;
```
PLeVasseur, line 63: `UStatus` contains a `UCode`, right?

Can we have this look like:
```suggestion
	    utils::Expected<v1::UMessage, v1::UStatus>;
```
instead?

I am just thinking this could be a bit of a cognitive burden on having to check the `variant`. WDYT?

include/up-cpp/datamodel/builder/UMessage.h
```diff
@@ -0,0 +1,258 @@
...
+	/// but a builder where withId() has been called will always produce the ID
+	/// used in that call.
+	///
+	/// @note In general, use of this method represents a special case. It
+	///       does not need to be used in most scenarios as a unique
+	///       message ID will be generated automatically for every built
+	///       message.
+	///
+	/// @post This UMessageBuilder will only generate messages with their
+	///       ID set to the value passed to this method.
```
PLeVasseur, line 117: Could maybe leave a further blurb warning about this method, given that the UUID contains not only the `rand_b` portion for uniquely identifying the uE, but also contains the timestamp & counter.

Just trying to think of more flashing neon signs to put above this function say -- "hey you probably didn't mean to use this!"

include/up-cpp/datamodel/validator/Uuid.h
```diff
@@ -0,0 +1,120 @@
...
+// @brief Collection of functions for validating and inspecting UUIDs.
+namespace uprotocol::datamodel::validator::uuid {
+
+enum class Reason {
+	/// @brief The UUID could not be interpreted as a uP v8 UUID
+	WRONG_VERSION,
+	/// @brief The UUID variant is not a supported variant
+	UNSUPPORTED_VARIANT,
+	/// @brief The time in the UUID is in the future
+	FROM_THE_FUTURE,
```
PLeVasseur, line 32: :sunglasses:

test/coverage/communication/NotificationSinkTest.cpp
PLeVasseur, line 1: Loving that you set up all the test fixtures and test files! :heart:

include/up-cpp/README.md
gregmedd: 99.9% of them, but it's a fair point

include/up-cpp/communication/NotificationSource.h
gregmedd: Yes, but more specifically [getDefaultSource()](https://github.com/eclipse-uprotocol/up-cpp/pull/105/files#diff-a36088a3d7aefa487d703f85584902256a8d27e3daec98f6cdaa6edced6d44a6R119)

include/up-cpp/communication/RpcClient.h
gregmedd: I have mixed feelings on it. Most likely it'll turn a call like this:
```
RpcClient foo(transport, methodUri, UPRIORITY_CS4, 100ms);
```
and turn it into this:
```
RpcClient foo({.transport=transport, .method=methodUri, .priority=UPRIORITY_CS4, .ttl=100ms});
```
I'm not sure that significantly improves readability in this case.

include/up-cpp/communication/RpcClient.h
gregmedd: commstatus is not defined as a UCode. It's just an arbitrary uint32

gregmedd: :neutral_face: 

It changed.

gregmedd: TBH I'm not sure why commstatus isn't just a whole UStatus. The non-enum fields are all optional, so it wouldn't significantly increase message sizes.

In any case, I'm going to go change that now :confused:

gregmedd: I'm actually still going to keep the types a little separate so it is possible to tell the difference between a status returned locally and a status received in a message. But it'll be a v1::UStatus and a v1::UCode now.

include/up-cpp/datamodel/builder/UMessage.h
gregmedd: I'm thinking it might be safer to just remove. The message builder will be inserting valid ones automatically, and if there is a need to replace them (e.g. for testing or for message playback), that can be done after calling build()

include/up-cpp/communication/RpcClient.h
PLeVasseur: > :neutral_face: 
> 
> It changed.

Caught me off guard too when I went looking 😅

include/up-cpp/communication/RpcClient.h
PLeVasseur: Ah I see what you mean. Fair point.

gregmedd:
The API has already been reviewed and approved in PR #99. I'm going to go ahead and merge this so that implementation work can start.

PR #123, opened 2024-05-23
Implementation for CallbackConnection.h

### LINKED ISSUE #109: Consider switching to C++20
We have received feedback that many consumers of this library have already migrated to C++20. We should evaluate if it is safe to roll forward so that newer features can be used in the interface and implementation.

### LINKED ISSUE #112: Implement and test utils::callbacks
@gregmedd already has most of the implementation, so it just needs to be copied into the appropriate files and tested.

gregmedd:
Provides both the implementation and a test for the callback connection utility.

~~Posted in draft status until some accompanying documentation describing the design for this module can be added to the PR.~~

~~Note that the current change would switch the repo to C++20, pending confirmation that we are safe to do so. This is in part because the code in this PR heavily relies on C++20 features. While it is possible to rework to use C++17 features instead, some aspects of the implementation will be more complicated.~~

resolves #112

### DIFF
--- a/docs/design/utils/CallbackConnection.md
+++ b/docs/design/utils/CallbackConnection.md
@@ -1,0 +1,382 @@
+# Callback Connection: Scope-managed callbacks as connected pairs of handles
+
+The Callback Connection module aims to improve safety around the use of
+callback functions. It provides each participant in a callback relationship,
+the _Caller(s)_ and _Callee_, a handle representing that connection. When
+either handle is discarded, the connection is broken with no manual
+intervention (i.e. unregistering the callback) required.
+
+## Background
+One of the core philosphies behind modern C++ is _Resource Acquisition is
+Intialization_, or **RAII**, wherein the core tenets are:
+
+* Resource acquisition occurs at object creation
+* Resource release occurs at destruction
+
+Which is to say, resources are only held while the owning object exists.
+Combined with C++'s destroy-at-scope-exit object lifecycle management, this
+provides a critical safety advantage over separate acquisition and release
+semantics. Regardless of the code path taken to exit a scope, the required
+resource release will occur.
+
+As the C++ Core Guidelines puts it: [Don't leak any resources][cpp-core-raii].
+
+This philosophy is applied throughout the C++ Standard Library. For example,
+[`std::unique_ptr`][cpp-ref-uniq] replaces paired `new`/`delete` calls with
+a scoped object. Discard the `unique_ptr` and the memory resource it holds
+is released.
+
+## Design
+
+The Callback Connection module provides an RAII model for representing a
+connection between an owner and a user of a callback function. This function
+serves as the resource, with the owner of this resource (the _callee_) creating
+a connection to a non-owning _caller_ that is expected to utilize the callback
+function in the future. Each participant in this connection is provided a
+handle that, when discarded, automatically severs the connection.
+
+```mermaid
+erDiagram
+    "Callee Handle" ||--|| "Callback" : owns
+    "Callee Handle" ||..|| "Connection" : "weakly references"
+    "Connection" ||..|| "Callback" : "weakly references"
+    "Caller Handle" }|--|| "Connection": shares
+    "Caller Handle" }|..|| "Callback" : "calls (via Connection)"
+```
+
+The combination of shared, owned, and weak reference relationships allows for
+either end to break the connection. In our implementation, we will use
+[`std::shared_ptr`][cpp-ref-shared] and [`std::weak_ptr`][cpp-ref-weak] to form
+these relationships.
+
+The _Caller Handle_ breaks the connection simply by dropping its instance of
+the connection `shared_ptr` - once no more _Caller Handles_ hold a `shared_ptr`
+to the connection, the _Callee Handle's_ `weak_pointer` will become invalid
+(so long as it was not [locked][cpp-ref-weak-lock]).
+
+The _Callee Handle_ must take slightly more care when severing the connection.
+This is because the callback function may access external resources that may
+become unavailable once the _Callee Handle_ believes the conneciton is broken.
+As such, when the _Callee Handle_ attempts to sever the connection, it must
+set a flag in the _Connection_ to prevent any new callbacks from executing,
+then wait until all currently execting callbacks have completed. Additional
+details of this behavior are discussed in the section on thread safety below.
+
+### Interface and Class Composition
+
+This module will provide interfaces through three objects:
+
+* `Connection`
+* `CallerHandle`
+* `CalleeHandle`
+
+```mermaid
+classDiagram
+    class CallerHandle {
+        -shared_ptr~Connection~ connection_
+        +operator()(...)
+        +reset()
+        +operator bool()
+        +operator==(shared_ptr~Connection~)
+    }
+    class Connection {
+      -weak_ptr~function~ callback_
+      +establish(...)$
+      +operator()(...)
+      -sever()
+    }
+    class CalleeHandle {
+        -weak_ptr~Connection~ connection_
+        -shared_ptr~function~ callback_
+        -optional~cleanup_function~ cleanup_
+        +reset()
+        +operator bool()
+    }
+
+    CalleeHandle *.. Connection
+    CallerHandle *-- Connection
+```
+
+The `Connection` object represents the connection between the two handles.
+It is templated on both the return type and argument types of the callback
+function. Pairs of connected handles are created using its static `establish()`
+method:
+
+```
+using Callback = std::function<ReturnT(Args...)>;
+using Cleanup = std::function<void(shared_ptr<Connection>)>;
+using ConnectedPair = std::tuple<CalleeHandle, CallerHandle>;
+
+static ConnectedPair establish(Callback&& cb, std::optional<Cleanup>&& = {});
+```
+
+When establishing a connection, a callback function must be provided.
+Additionally, an optional cleanup function can be provided that will be called
+when the connection is broken from the _callee_ side. This will be provided
+a value that can be checked against any held `CallerHandle` instances in a
+container, such as might be found in a callback registry for a messaging
+service.
+
+As `CalleeHandle` owns the callback function, it is non-copyable. However, it
+can be moved. The holder of a `CalleeHandle` can break the connection in one of
+two ways: discarding the object, or calling `reset()` which restores
+`CalleeHandle` to its default, non-connected state.
+
+`CallerHandle` can be copied as it might need to be shared in the context of
+providing a service (e.g. when serving callbacks using a thread pool). It
+provides a function call operator that will check the status of the connection
+then forward all arguments to the callback function. The holder(s) of a
+`CalleeHandle` can break the connection by discarding or calling `reset()` on
+ _all_ instances of the handle.
+
+Both `CalleeHandle` and `CallerHandle` provide conversion operators to `bool`
+that will return `true` if the connection is still active or `false` if the
+connection has been severed.
+
+### Thread Safety and Parallel Access
+
+In many systems using callback functions, it is likely that the _Caller_ and
+_Callee_ will operate from separate execution contexts. Further, some patterns
+may allow for multiple simultaneous invokations of the same callback function.
+
+On the _Caller_ side, each separate copy of the `CallerHandle` instance will
+have a copy of the shared pointer to the `Connection` instance. Separate copies
+of a `shared_ptr` are [guaranteed to be thread safe][cpp-ref-shared] - the
+shared state is accessed and updated atomically. No additional locking is
+required here.
+
+In the middle of both handles is the `Connection` object. A single instance
+is shared by both the _Callee Handle_ and all _Caller Handle_ copies. This
+holds a `std::weak_ptr` to the callback function. Since this is a single
+instance on the weak pointer object, manipulation of that weak pointer must
+be guarded against concurrent access. This protection is implemented with
+`std::mutex`, `std::condition_variable`, and `std::scoped_lock`:
+
+```mermaid
+sequenceDiagram
+    participant C as callback
+    box Connection
+        participant W as weak_ptr
+        participant M as mutex
+        participant L as scoped_lock
+        participant Op as operator()
+    end
+    box CallerHandle
+        participant CHOp as operator()
+    end
+    actor Events
+
+    par
+        Events --)+ CHOp: event
+        CHOp ->>+ Op: 
+        Op ->>+ L: lock(m)
+        L ->>+ M: lock()
+    and
+        Events --)+ CHOp: event
+        CHOp ->>+ Op: 
+        Op ->>+ L: lock(m)
+        L ->>+ M: lock()
+    end
+
+    M ->>- L: 
+    L ->>- Op: 
+    Op ->>+ W: lock()
+    W ->>- Op: shared_ptr copy
+    Op ->>+ L: unlock()
+    L ->>+ M: unlock()
+    M ->>- L: 
+    L ->>- Op: 
+
+    M ->>- L: 
+    L ->>- Op: 
+
+    par
+        Op ->>+ C: 
+        C ->> C: working
+        C ->>- Op: 
+        Op ->>- CHOp: 
+    and
+        Op ->>+ W: lock()
+        W ->>- Op: shared_ptr copy
+        Op ->>+ L: unlock()
+        L ->>+ M: unlock()
+        M ->>- L: 
+        L ->>- Op: 
+        Op ->>+ C: 
+        C ->> C: working
+        C ->>- Op: 
+        Op ->>- CHOp: 
+    end
+```
+
+Last is the _Callee_ side of the connection. The `CalleeHandle` for a given
+connection cannot be copied and is expected to have a single owner. It is not
+shared. As such, its implementation does not need to directly address
+concurrency or locking.
+
+#### Severing the Connection with Active Calls
+
+There is one case where additional care is required around locking: severing
+the connection from the _Callee_ side while there are active calls to the
+callback function. We must ensure that resources used by the callbacks are
+not released while they are still executing.
+
+The owner of the _callee handle_ will, in most cases, have initialized the
+resources the callback depends on before connecting the callback. When this is
+the case, C++ will delete the resources in the reverse of the initialization
+order. Consider this code:
+
+```
+int countEventRate() {
+    atomic<int> eventsSeen{0};
+    auto handle = listenForEvents([&eventsSeen]() { ++eventsSeen; });
+    std::this_thread::sleep_for(1s);
+    return eventsSeen;
+}
+```
+
+The `handle` will be deleted before `eventsSeen`. In order to achieve our goal
+of keeping resources alive until callbacks complete, the destruction of the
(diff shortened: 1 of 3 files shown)


test/coverage/utils/CallbackConnectionTest.cpp
gregmedd, line 1: This test could really use some comments explaining what it is doing and why it's running specific checks.

include/up-cpp/utils/CallbackConnection.h
gregmedd, line 1: Add a design doc file that at least explains the locking behavior around `sever()` with a sequence diagram.

CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ find_package(up-core-api REQUIRED)
 # This is the root CMakeLists.txt file; We can set project wide settings here
 # TODO: Is this needed?
 if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
-	set(CMAKE_CXX_STANDARD 17)
+	set(CMAKE_CXX_STANDARD 20)
```
gregmedd, line 25: Should the changes to switch to C++20 move to a different PR? Or are we OK reviewing them here since the implementation of the callback utils requires it right now.

gregmedd:
The CI issues here are caused by a recent change in up-spec. A PR to fix it is pending: https://github.com/eclipse-uprotocol/up-spec/pull/161

.github/workflows/ci_conan_profile
stevenhartley, line 1: Why do we need this? Is this how CCI handles testing building for different configurations? How would we test for the arm/arch64, etc...

gregmedd: We need this to select C++20 instead of the default C++17 in the github build environment.

Profiles are how CCI (and conan in general) handle different build configurations and targets. We could add more profiles to test cross-compilation if we wanted. I believe github also has armv8 runners, so we could take the cross-compiled build then run the tests on a real arm core.

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -164,33 +202,48 @@ struct [[nodiscard]] CalleeHandle {
...
+	void reset() {
+		callback_.reset();
+		// Forces us to wait until all active callbacks have returned
+		if (connection_) {
+			connection_->sever();
+		}
+		// Optionally, let someone know they need to clean up
+		if (cleanup_) {
+			(*cleanup_)(connection_);
+		}
```
gregmedd, line 245: Unit test doesn't cover the true branch of this check.

gregmedd:
Pulling back to draft status so I can split the commits.

CMakeLists.txt
gregmedd: The change to C++20 has been removed from this PR. New PRs to follow.

gregmedd:
PR is ready for review again. C++20 changes have been split out and will be put in a separate PR.

debruce (commented):
Given that there is a unit test, I would presume that technical validation of this design has occurred. Of course I will still go through it on a technical basis to review. However, up front, it seems reasonable to ask why this is needed at all, since it isn't a pattern I'm aware of being discussed for std inclusion, nor have I seen the idea in Boost. I did a working demo that does have the need for this.
It seems that registerListener() is the source of the need. The need for callbacks occur in rpc server, subscriber, and notification server, but not in rpc client since that is being handled by a future. In each needed case, there is deregistration on the Zenoh-cpp API or its alternative, and deregistration from possible thread pools or destruction of a thread pool, etc. It seems that the intent of this design is that registerListener() interacts with Zenoh-cpp, and the destruction of the thread pool or other resources has to be handled has a call path from higher level code. If the design were to have a subscriber object, for example, that implements Zenoh-cpp reg/dereg, and the thread pool, and allows for derivation or shared_ptr holding to bond the callback behavior to the subscriber object, then there is no possibility for things to get out-of-sync between listener registration and other resource deallocation.
Having said that, I think it's instructive to look at a hypothetical case of resources to be deallocated, because the simple callback case (for subscriber example) can just be a std::function<void (UMessage)>. A unit test for this feature should only have to deal with such a signature. Beyond a simple unit test, you may want to have the callback be able to record information that its seeing into a data structure for aggregate analysis on exit, or whatever. For a case like this, the top level object contains the aggregating info data structure, the callback as a callable method within this class (or a lambda function customizing to the class method), and a unique_ptr or suitable ref to the subscriber object. The parent object's life/death cycle contains the subscriber object's life/death cycle, and the subscriber object's life/death cycle controls Zenoh-cpp (or other) registration/deregistration under the hood. There should not be a registerListener() method. Its roile should be implicitly filled by the construction and destruction of the subscriber API object.
Having said this, I wish I could have gotten these basic points across a month ago.

debruce:
In a follow-up to the above comment, I would repeat Steven's question about 'why do we need this'. Expecting C++20 to cover for a basic design issue seems risky and unwise. RAII suuggests a simpler path than this compositional approach extending up to the API.

docs/design/utils/CallbackConnection.md
gregmedd, line 1: Remove references to C++20 from this design. That can be saved for a future PR.

.github/workflows/ci_conan_profile
gregmedd: This has been removed. It will be handled in a separate PR.

gregmedd:
> In a follow-up to the above comment, I would repeat Steven's question about 'why do we need this'.

@debruce - I am assuming you are talking about [this comment](https://github.com/eclipse-uprotocol/up-cpp/pull/123#discussion_r1613766594). Steven was asking why I had added a conan profile file to the CI. However, that has since been removed and this PR no longer involves any C++20 features.

> Expecting C++20 to cover for a basic design issue seems risky and unwise. RAII suuggests a simpler path than this compositional approach extending up to the API.

I don't know where you're coming from on this. I don't expect C++20 to _cover for a basic design issue_. It simply provides library features that implement the same functionality with less code on our end.

gregmedd:
@debruce - I'm going to copy something I wrote near the end of my reply up to the top because I feel it might be the core of the misunderstanding here:

`Subscriber` is a layer 2 API and `UTransport` is a layer 1 API. It must be possible to use `UTransport` without using `Subscriber`. We have some use cases that will operate entirely at L1. We _cannot_ remove `registerListener()` from `UTransport`.

----

> However, up front, it seems reasonable to ask why this is needed at all, since it isn't a pattern I'm aware of being discussed for std inclusion, nor have I seen the idea in Boost.

It is needed to provide an alternative to a separate `unregisterListener()` interface. That risks leaving callbacks registered even after the code that initially registered the callback has long since returned and discarded its data. It is fairly common in the standard library to provide interfaces where the destruction of an object cleans up some state in order to avoid requiring a separate call to do so. See `std::lock_guard` and `std::unique_ptr`.

As for this exact concept being included in the std library or Boost: lots of code exists in the world that is not included in those places. It does not invalidate that other code. In the case of this particular model (representing a callback registration as a connection with discardable handles), I have seen it used very successfully in several code bases.

> I did a working demo that does have the need for this. It seems that registerListener() is the source of the need. The need for callbacks occur in rpc server, subscriber, and notification server, but not in rpc client since that is being handled by a future.

`registerListener()` and the issues caused by requiring a separate `unregisterListener()`, yes. That is the source of the need. Additionally, this provides a mechanism for the caller end to signal to the callee end that it is no longer using the callback (e.g. in the case of RPC response handlers, or when the transport has shut down).

> In each needed case, there is deregistration on the Zenoh-cpp API or its alternative, and deregistration from possible thread pools or destruction of a thread pool, etc.

I'm not sure what zenoh-cpp has to do with this at all.

> It seems that the intent of this design is that registerListener() interacts with Zenoh-cpp,

The intent of this design is that, when a callback is registered, a contract is formed between the calling and called parties. They are each given a handle representing that contract and, so long as that handle is held, the contract remains valid.

> and the destruction of the thread pool or other resources has to be handled has a call path from higher level code.

I don't follow.

> If the design were to have a subscriber object, for example, that implements Zenoh-cpp reg/dereg, and the thread pool, and allows for derivation or shared_ptr holding to bond the callback behavior to the subscriber object, then there is no possibility for things to get out-of-sync between listener registration and other resource deallocation.

TBH what you have described here does not sound too dissimilar from what is implemented here. However, you have to consider that the "subscriber" (callee) is potentially executing in a separate context from the transport, and callbacks could potentially be executing when the subscriber tries to deregister. Additional care is required to ensure that callbacks are allowed to return before the contract is broken.

> Having said that, I think it's instructive to look at a hypothetical case of resources to be deallocated, because the simple callback case (for subscriber example) can just be a std::function<void (UMessage)>.

Don't forget that the `std::function<void(UMessage)>` could be a lambda with captures.

> A unit test for this feature should only have to deal with such a signature.

Sure, but not all of the places we use this module have the exact same callback signature (see `RpcServer`). It adds very little actual complexity to make it generic.

> Beyond a simple unit test, you may want to have the callback be able to record information that its seeing into a data structure for aggregate analysis on exit, or whatever. For a case like this, the top level object contains the aggregating info data structure, the callback as a callable method within this class (or a lambda function customizing to the class method), and a unique_ptr or suitable ref to the subscriber object.

The callback _can_ access other data to record aggregate information. The callback can access class members if it is a class method, or it can access anything it captures if it is a lambda. Alternatively, interfaces wanting to allow callbacks to return a value (see `RpcServer`) can do so by setting the `RT` template parameter to some non-void value.

Providing a mechanism for callbacks to aggregate information goes beyond the scope of the code in this review.

> The parent object's life/death cycle contains the subscriber object's life/death cycle, and the subscriber object's life/death cycle controls Zenoh-cpp (or other) registration/deregistration under the hood. There should not be a registerListener() method. Its roile should be implicitly filled by the construction and destruction of the subscriber API object. Having said this, I wish I could have gotten these basic points across a month ago.

I feel like this has been said many times, but `Subscriber` is a L2 API and `UTransport` is a L1 API. It must be possible to use `UTransport` without using `Subscriber`. We have some use cases that will operate entirely at L1. We _cannot_ remove `registerListener()` from `UTransport`.

However, most uE code _will_ use L2 interfaces like `Subscriber`, and in that case the entire `registerListener()` thing is done for them by the L2 object.

gregmedd:
@debruce - I have a request going forward: Can you please make all comments from the _files changed_ tab of this PR, highlighting the exact lines you are commenting on? I would like to keep this review focused on whether the code is ready to be merged. The use of this model for representing callbacks has already been litigated several times.

stevenhartley (approved):
LGTM, please get +1 from another C++ developer

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -118,79 +150,151 @@ struct [[nodiscard]] Connection {
...
+				callback = callback_;
+			}
+
+			if (auto locked_cb = callback.lock(); locked_cb) {
+				(*locked_cb)(std::forward<Args>(args)...);
+			}
+
+			notify_if_expired(std::move(callback));
+		}
+	}
```
gregmedd, line 168: So I just figured out a way to combine both `operator()` versions into a single implementation. Any interest in me adding that to this PR? I think it would probably be better in the long run; it avoids two places where possible fixes could be needed.

billpittman: I suggest putting that change in another PR, this one is close to submission.

gregmedd: I'll hold off on this unless there are other changes requested.

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -200,14 +304,97 @@ struct [[nodiscard]] CalleeHandle {
...
+	///     * True if the connection is valid (i.e. at least one other
+	///       reference to the connection still exists).
+	///     * False if the connection has been broken (i.e. This handle has
+	///       been reset/moved, or all other references to the connection
+	///       have been discarded)
+	explicit operator bool() const { return connection_ && (*connection_); }
+
+	/// @throws BadCallerAccess if this handle has been default constructed OR
+	///         reset() has left it without a valid conneciton pointer.
+	auto operator()(Args&&... args) {
```
MarkStokes71, line 363: These uses of auto are non-compliant to Autosar.  Here's a reference: Rule A7-1-5

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -200,14 +304,97 @@ struct [[nodiscard]] CalleeHandle {
...
+			throw BadCallerAccess(
+			    "Cannot call a CallerHandle that is in the reset state");
+		}
+		return (*connection_)(std::forward<Args>(args)...);
+	}
+
+	/// @brief Comparison based on connection pointer value
+	///
+	/// Allows `CallerHandle` to be stored in a container that sorts on value.
+	auto operator<(const CallerHandle& other) const {
```
MarkStokes71, line 383: Is the point of this to allow sorted lists by caller address?  This doesn't seem to provide any value.

test/coverage/utils/CallbackConnectionTest.cpp
```diff
@@ -23,16 +28,469 @@ class TestFixture : public testing::Test {
...
+			// Callable dropped here
+		}
+		EXPECT_FALSE(handle_outside);
+	}
+
+	// If multiple copies of the callable end are held, dropping one doesn't
+	// break the connection.
+	{
+		auto [handle, callable] = getPair();
+		auto other_callable = callable;
```
MarkStokes71, line 145: Point of note, test code does not need to be Autosar compliant.

MarkStokes71 (commented):
Nice work.

include/up-cpp/utils/CallbackConnection.h
gregmedd: It allows for handles to be stored in a `std::map` if needed for reverse lookups during the cleanup operation

include/up-cpp/utils/CallbackConnection.h
gregmedd: Autosar is not, to my knowledge, a requirement of up-cpp

test/coverage/utils/CallbackConnectionTest.cpp
gregmedd: Per Autosar A7-1-5:

> The auto specifier shall not be used apart from following cases: (1) to declare
that a variable has the same type as return type of a function call...

This use of `auto` is Autosar compliant.

include/up-cpp/utils/CallbackConnection.h
gregmedd: Reviewing Autosar A7-1-5, this use of `auto` would be allowed with a trailing return type to aid in type deduction:

> The auto specifier shall not be used apart from following cases: ... (4) to declare a function template using trailing return type syntax.

Based on that, a compliant fix would be:
```suggestion
	auto operator()(Args&&... args) -> decltype(*connection_)(std::forward<Args>(args)...)) {
```

(yes, the examples provided for a valid case 4 in Autosar A7-1-5 do use `decltype()` like that)

Personally, I find using a trailing return type in this context redundant - the exact same type deduction will occur regardless since the code that generates the return is copied twice.

Incidentally, Autosar has been merged with MISRA for C++17 and up based on what I've read, so the referenced rules are likely out of date for projects using anything after C++14.

include/up-cpp/utils/CallbackConnection.h
gregmedd: Change committed. I'll push it once I've addressed more of the comments.

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -86,7 +97,10 @@ struct [[nodiscard]] Connection {
 	///       not been reset).
 	///     * False if the connection has been broken (i.e. the Handle was
 	///       discarded or reset).
-	explicit operator bool() const;
+	explicit operator bool() const {
```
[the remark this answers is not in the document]
gregmedd: Added `isConnected()` methods.

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -118,79 +150,151 @@ struct [[nodiscard]] Connection {
...
+struct BadConnection : public std::runtime_error {
+	template <typename... Args>
+	BadConnection(Args&&... args)
+	    : std::runtime_error(std::forward<Args>(args)...) {}
+};
+
 template <typename RT, typename... Args>
 struct [[nodiscard]] CalleeHandle {
 	using Conn = Connection<RT, Args...>;
 	using Callback = typename Conn::Callback;
```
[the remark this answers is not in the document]
gregmedd: Removed. It does mean that the `typename` keyword becomes necessary, so instead of `shared_ptr<Callback>` it will read `shared_ptr<typename Conn::Callback>`.

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -118,79 +150,151 @@ struct [[nodiscard]] Connection {
...
+/// reason.
+struct BadConnection : public std::runtime_error {
+	template <typename... Args>
+	BadConnection(Args&&... args)
+	    : std::runtime_error(std::forward<Args>(args)...) {}
+};
+
 template <typename RT, typename... Args>
 struct [[nodiscard]] CalleeHandle {
 	using Conn = Connection<RT, Args...>;
```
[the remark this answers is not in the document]
gregmedd: I agree - I don't like them. But it's not possible to have `using Connection = Connection<RT, Args...>` and putting the full templating list everywhere, in my opinion, makes the code harder to read. So given the list of not great choices, this isn't the worst option for readability.

test/coverage/utils/CallbackConnectionTest.cpp
gregmedd: (I know I submitted another comment here, but it's not showing up so I'm trying again)

I must have needed more sleep when I wrote that earlier comment. The letter of the requirement in A7-1-5 would say that `auto other_callable = callable` is not allowed. However, the _meaning_ of exception 1 effectively applies here. There is no intervening code between the line before where a value is assigned auto from a function return and where that value is immediately copied to another value. The type inferences operate the same in both cases. There would be little to no value in the first use of `auto` it the second use were not allowed.

include/up-cpp/utils/CallbackConnection.h
gregmedd: I think where we are at is that we need to see if we can get our hands on MISRA C++ 2023 to check the latest guidance for C++17 and C++20. We will also be asking for static analysis to be run against this code base. For the moment, I'd like to defer these changes until we have the appropriate references. There are several P0 and P1 issues waiting on the completion of this module. We will make a clean-up pass when possible to address MISRA/AUTOSAR compliance.

test/coverage/utils/CallbackConnectionTest.cpp
gregmedd: Tracking request in #154

include/up-cpp/utils/CallbackConnection.h
gregmedd: Tracking request in #154

gregmedd:
All open threads have either been resolved or moved to an issue to track for future work. Once an approval is added (@MarkStokes71, if you are ok with the state of this PR), this PR can be merged.

MarkStokes71 (approved):

PR #124, opened 2024-05-23
Header implementation to mock C++23 std::expected.

debruce:
This PR will deliver a mocking of the C++23 std::expected feature minus the complexities of reference type overloads. User behavior should be the same, but performance for the returning of large expected objects could be worse due to copies.

### DIFF
--- a/include/up-cpp/utils/Expected.h
+++ b/include/up-cpp/utils/Expected.h
@@ -12,8 +12,10 @@
 #ifndef UP_CPP_UTILS_EXPECTED_H
 #define UP_CPP_UTILS_EXPECTED_H
 
-#include <exception>
+#include <stdexcept>
 #include <type_traits>
+#include <utility>
+#include <variant>
 
 namespace uprotocol::utils {
 
@@ -27,72 +29,100 @@ static_assert(!__has_cpp_attribute(__cpp_lib_expected),
 ///          https://en.cppreference.com/w/cpp/utility/expected
 ///          No further documentation is provided in this file.
 /// @{
-template <typename E>
-struct BadExpectedAccess;
+struct BadExpectedAccess : public std::runtime_error {
+	template <typename... Args>
+	BadExpectedAccess(Args&&... args)
+	    : std::runtime_error(std::forward<Args>(args)...) {}
+};
 
-template <>
-struct BadExpectedAccess<void> : public std::exception {};
-
+/// @brief Required tagging type for cases where expected and unexpected type
+/// are identical.
 template <typename E>
-struct BadExpectedAccess : public BadExpectedAccess<void> {};
+class Unexpected {
+public:
+	constexpr Unexpected(const Unexpected&) = default;
+	constexpr Unexpected(Unexpected&&) = default;
+
+	constexpr explicit Unexpected(const E& rhs) : storage_(rhs) {}
+	constexpr explicit Unexpected(E&& rhs) : storage_(std::move(rhs)) {}
+
+	constexpr const E& error() const& noexcept { return storage_; }
+	constexpr E&& error() && noexcept { return std::move(storage_); }
+
+private:
+	E storage_;
+};
 
 /// @brief A stripped-down version of std::expected from C++23.
 template <typename T, typename E>
-struct Expected {
-	using value_type = T;
-	using error_type = E;
+class Expected {
+public:
+	template <typename... Args>
+	constexpr Expected(Args&&... args)
+	    : storage_(std::forward<Args>(args)...) {}
 
-	constexpr Expected();
-	constexpr Expected(const Expected& other);
-	constexpr Expected(Expected&& other) noexcept;
+	constexpr Expected(const Expected&) = default;
+	constexpr Expected(Expected&&) = default;
 
-	constexpr explicit Expected(T&& v);
-	constexpr explicit Expected(E&& e);
+	constexpr bool has_value() const noexcept {
+		return std::holds_alternative<T>(storage_);
+	}
 
-	Expected& operator=(const Expected& other);
-	Expected& operator=(Expected&& other);
+	constexpr explicit operator bool() const noexcept {
+		return std::holds_alternative<T>(storage_);
+	}
 
-	Expected& operator=(const T& other);
-	Expected& operator=(T&& other);
+	template <class X>
+	constexpr T value_or(X&& v) const& noexcept {
+		return has_value() ? std::get<T>(storage_)
+		                   : static_cast<T>(std::forward<X>(v));
+	}
 
-	Expected& operator=(const E& other);
-	Expected& operator=(E&& other);
+	constexpr const T& value() const& {
+		if (!has_value())
+			throw BadExpectedAccess(
+			    "Attempt to access value() when unexpected.");
+		return std::get<T>(storage_);
+	}
 
-	constexpr explicit operator bool() const noexcept;
-	constexpr bool has_value() const noexcept;
+	constexpr T value() && {
+		if (!has_value())
+			throw BadExpectedAccess(
+			    "Attempt to access value() when unexpected.");
+		return std::move(std::get<T>(storage_));
+	}
 
-	constexpr T& value() &;
-	constexpr const T& value() const&;
-	constexpr T&& value() &&;
-	constexpr const T&& value() const&&;
+	constexpr const E& error() const& {
+		if (has_value())
+			throw BadExpectedAccess(
+			    "Attempt to access error() when not unexpected.");
+		return std::get<Unexpected<E>>(storage_).error();
+	}
 
-	constexpr const E& error() const& noexcept;
-	constexpr E& error() & noexcept;
-	constexpr const E&& error() const&& noexcept;
-	constexpr E&& error() && noexcept;
+	constexpr E error() && {
+		if (has_value())
+			throw BadExpectedAccess(
+			    "Attempt to access error() when not unexpected.");
+		return std::move(std::get<Unexpected<E>>(storage_)).error();
+	}
 
-	constexpr T value_or(T&& default_value) const&;
-	constexpr T value_or(T&& default_value) &&;
+	constexpr const T& operator*() const {
+		if (!has_value())
+			throw BadExpectedAccess(
+			    "Attempt to dereference expected value when unexpected.");
+		return std::get<T>(storage_);
+	}
 
-	template <typename F>
-	constexpr auto and_then(F&& f) &;
-	template <typename F>
-	constexpr auto and_then(F&& f) const&;
-	template <typename F>
-	constexpr auto and_then(F&& f) &&;
-	template <typename F>
-	constexpr auto and_then(F&& f) const&&;
-
-	template <typename F>
-	constexpr auto or_else(F&& f) &;
-	template <typename F>
-	constexpr auto or_else(F&& f) const&;
-	template <typename F>
-	constexpr auto or_else(F&& f) &&;
-	template <typename F>
-	constexpr auto or_else(F&& f) const&&;
+	constexpr const T* operator->() const {
+		if (!has_value())
+			throw BadExpectedAccess(
+			    "Attempt to dereference expected pointer when unexpected.");
+		return &std::get<T>(storage_);
+	}
 
 private:
+	std::variant<T, Unexpected<E>> storage_;
+
 	static_assert(!std::is_void_v<T>,
 	              "We don't allow T==void (unlike std::expected)");
 
@@ -101,6 +131,7 @@ private:
 	        !std::is_reference_v<T>,
 	    "Expected requires T to meet the C++ 'destructable' requirement");
 };
+
 /// @}
 
 }  // namespace uprotocol::utils
--- a/test/coverage/utils/ExpectedTest.cpp
+++ b/test/coverage/utils/ExpectedTest.cpp
@@ -12,9 +12,15 @@
 #include <gtest/gtest.h>
 #include <up-cpp/utils/Expected.h>
 
+#include <random>
+
 namespace {
 
-class TestFixture : public testing::Test {
+using uprotocol::utils::BadExpectedAccess;
+using uprotocol::utils::Expected;
+using uprotocol::utils::Unexpected;
+
+class ExpectedTest : public testing::Test {
 protected:
 	// Run once per TEST_F.
 	// Used to set up clean environments per test.
@@ -23,8 +29,8 @@ protected:
 
 	// Run once per execution of the test application.
 	// Used for setup of all tests. Has access to this instance.
-	TestFixture() = default;
-	~TestFixture() = default;
+	ExpectedTest() = default;
+	~ExpectedTest() = default;
 
 	// Run once per execution of the test application.
 	// Used only for global setup outside of tests.
@@ -32,7 +38,260 @@ protected:
 	static void TearDownTestSuite() {}
 };
 
-// TODO replace
-TEST_F(TestFixture, SomeTestName) {}
+int get_rand() {
+	static std::random_device rd;
+	static std::mt19937 mt(rd());
+	static std::uniform_int_distribution<int> dist(0, 1 << 30);
+	return dist(mt);
+}
+
+TEST_F(ExpectedTest, ExpectScalarScalar) {
+	auto sample = get_rand();
+	auto expected = Expected<int, int>(sample);
+	EXPECT_TRUE(bool(expected));
+	EXPECT_TRUE(expected.has_value());
+	EXPECT_EQ(sample, expected.value());
+	EXPECT_EQ(sample, *expected);
+}
+
+TEST_F(ExpectedTest, UnexpectScalarScalar) {
+	int sample = get_rand();
+	auto expected = Expected<int, int>(Unexpected(sample));
+	EXPECT_FALSE(bool(expected));
+	EXPECT_FALSE(expected.has_value());
+	EXPECT_EQ(sample, expected.error());
+}
+
+TEST_F(ExpectedTest, ExpectScalar) {
+	auto sample = get_rand();
+	auto expected = Expected<int, std::string>(sample);
+	EXPECT_TRUE(bool(expected));
+	EXPECT_TRUE(expected.has_value());
+	EXPECT_EQ(sample, expected.value());
+	EXPECT_EQ(sample, *expected);
+}
+
+TEST_F(ExpectedTest, UnexpectScalar) {
+	int sample = get_rand();
+	auto expected = Expected<std::string, int>(Unexpected(sample));
+	EXPECT_FALSE(bool(expected));
+	EXPECT_FALSE(expected.has_value());
+	EXPECT_EQ(sample, expected.error());
+}
+
+TEST_F(ExpectedTest, UnexpectValueOr) {
+	int sample = get_rand();
+	auto expected =
+	    Expected<int, std::string>(Unexpected(std::string("hello")));
+	EXPECT_FALSE(bool(expected));
+	EXPECT_FALSE(expected.has_value());
+	EXPECT_EQ(sample, expected.value_or(sample));
+}
+
+struct Pair {
+	int x;
+	int y;
+
+	Pair(int x, int y) : x(x), y(y) {}
+};
+
+TEST_F(ExpectedTest, ExpectUnique) {
+	auto x = get_rand();
+	auto y = get_rand();
+	auto expected = Expected<std::unique_ptr<Pair>, std::string>(
+	    std::make_unique<Pair>(x, y));
+	EXPECT_TRUE(bool(expected));
+	EXPECT_TRUE(expected.has_value());
+	auto p = std::move(expected).value();
+	EXPECT_EQ(x, p->x);
+	EXPECT_EQ(y, p->y);
+}
+
+TEST_F(ExpectedTest, UnexpectUnique) {
+	auto x = get_rand();
+	auto y = get_rand();
+	auto expected = Expected<int, std::unique_ptr<Pair>>(
+	    Unexpected(std::make_unique<Pair>(x, y)));
+	EXPECT_FALSE(bool(expected));
+	EXPECT_FALSE(expected.has_value());
+	auto p = std::move(expected).error();
+	EXPECT_EQ(x, p->x);
+	EXPECT_EQ(y, p->y);
+}
+
+TEST_F(ExpectedTest, ExpectShared) {
+	auto x = get_rand();
+	auto y = get_rand();
+	auto expected = Expected<std::shared_ptr<Pair>, std::string>(
+	    std::make_shared<Pair>(x, y));
+	EXPECT_TRUE(bool(expected));
+	EXPECT_TRUE(expected.has_value());
+	EXPECT_EQ(x, expected.value()->x);
+	EXPECT_EQ(y, expected.value()->y);
+	EXPECT_EQ(x, (*expected)->x);
+	EXPECT_EQ(y, (*expected)->y);
+}
+
+TEST_F(ExpectedTest, UnexpectShared) {
+	auto x = get_rand();
+	auto y = get_rand();
+	auto expected = Expected<int, std::shared_ptr<Pair>>(
+	    Unexpected(std::make_shared<Pair>(x, y)));
+	EXPECT_FALSE(bool(expected));
+	EXPECT_FALSE(expected.has_value());
+	EXPECT_EQ(x, expected.error()->x);
+	EXPECT_EQ(y, expected.error()->y);
+}
+
+TEST_F(ExpectedTest, ExpectStruct) {
+	auto x = get_rand();
+	auto y = get_rand();
+	auto expected = Expected<Pair, std::string>(Pair(x, y));
+	EXPECT_TRUE(bool(expected));
+	EXPECT_TRUE(expected.has_value());
+	EXPECT_EQ(x, expected.value().x);
+	EXPECT_EQ(y, expected.value().y);
+	EXPECT_EQ(x, expected->x);
+	EXPECT_EQ(y, expected->y);
+}
+
+TEST_F(ExpectedTest, UnexpectStruct) {
+	auto x = get_rand();
+	auto y = get_rand();
+	auto expected = Expected<int, Pair>(Unexpected(Pair(x, y)));
+	EXPECT_FALSE(bool(expected));
+	EXPECT_FALSE(expected.has_value());
+	EXPECT_EQ(x, expected.error().x);
+	EXPECT_EQ(y, expected.error().y);
+}
+
+struct PairDestruct {
+	int x;
+	int y;
+	static int cd_count;
+
+	PairDestruct(int x, int y) : x(x), y(y) { cd_count++; }
+
+	PairDestruct(const PairDestruct& arg) : x(arg.x), y(arg.y) { cd_count++; }
+
+	~PairDestruct() { cd_count--; }
+};
+
+int PairDestruct::cd_count = 0;
+
+TEST_F(ExpectedTest, ExpectStructDestruct) {
+	PairDestruct::cd_count = 0;
+	{
+		auto x = get_rand();
+		auto y = get_rand();
+		auto expected = Expected<PairDestruct, std::string>(PairDestruct(x, y));
+		EXPECT_EQ(1, PairDestruct::cd_count);
+		EXPECT_TRUE(bool(expected));
+		EXPECT_TRUE(expected.has_value());
+		EXPECT_EQ(x, expected.value().x);
+		EXPECT_EQ(y, expected.value().y);
+		EXPECT_EQ(x, expected->x);
+		EXPECT_EQ(y, expected->y);
+	}
+	EXPECT_EQ(0, PairDestruct::cd_count);
+}
+
+TEST_F(ExpectedTest, UnexpectStructDestruct) {
+	PairDestruct::cd_count = 0;
+	{
+		auto x = get_rand();
+		auto y = get_rand();
+		auto expected =
+		    Expected<int, PairDestruct>(Unexpected(PairDestruct(x, y)));
+		EXPECT_EQ(1, PairDestruct::cd_count);
+		EXPECT_FALSE(bool(expected));
+		EXPECT_FALSE(expected.has_value());
+		EXPECT_EQ(x, expected.error().x);
+		EXPECT_EQ(y, expected.error().y);
+	}
+	EXPECT_EQ(0, PairDestruct::cd_count);
+}
+
+TEST_F(ExpectedTest, ExceptionValueCheckedWhenIsError) {
+	auto expected =
+	    Expected<int, std::string>(Unexpected(std::string("hello")));
+	EXPECT_THROW(
+	    {
+		    try {
+			    EXPECT_FALSE(bool(expected));
+			    EXPECT_FALSE(expected.has_value());
+			    auto value = expected.value();
+		    } catch (const BadExpectedAccess& ex) {
+			    EXPECT_STREQ("Attempt to access value() when unexpected.",
+			                 ex.what());
+			    throw;
+		    }
+	    },
+	    BadExpectedAccess);
+}
+
+TEST_F(ExpectedTest, ExceptionErrorCheckedWhenNotError) {
+	auto expected = Expected<int, std::string>(5);
+	EXPECT_THROW(
+	    {
+		    try {
+			    EXPECT_TRUE(bool(expected));
+			    EXPECT_TRUE(expected.has_value());
+			    auto err = expected.error();
+		    } catch (const BadExpectedAccess& ex) {
+			    EXPECT_STREQ("Attempt to access error() when not unexpected.",
+			                 ex.what());
+			    throw;
+		    }
+	    },
+	    BadExpectedAccess);
+}
+
+TEST_F(ExpectedTest, ExceptionDerefValueWhenUnexpected) {
+	auto expected =
+	    Expected<const Pair, std::string>(Unexpected(std::string("hello")));
+	EXPECT_THROW(
+	    {
+		    try {
+			    EXPECT_FALSE(bool(expected));
+			    EXPECT_FALSE(expected.has_value());
+			    const auto x = *expected;
+		    } catch (const BadExpectedAccess& ex) {
+			    EXPECT_STREQ(
+			        "Attempt to dereference expected value when unexpected.",
+			        ex.what());
+			    throw;
+		    }
+	    },
+	    BadExpectedAccess);
+}
+
+TEST_F(ExpectedTest, ExceptionDerefPtrWhenUnexpected) {
+	auto expected =
+	    Expected<Pair, std::string>(Unexpected(std::string("hello")));
+	EXPECT_THROW(
+	    {
+		    try {
+			    EXPECT_FALSE(bool(expected));
+			    EXPECT_FALSE(expected.has_value());
+			    auto x = expected->x;
+		    } catch (const BadExpectedAccess& ex) {
+			    EXPECT_STREQ(
+			        "Attempt to dereference expected pointer when unexpected.",
+			        ex.what());
+			    throw;
+		    }
+	    },
+	    BadExpectedAccess);
+}
 
 }  // namespace
+
+#ifdef OUT_OF_TREE
+
+int main(int argc, char* argv[]) {
+	::testing::InitGoogleTest(&argc, argv);
+	return RUN_ALL_TESTS();
+}
+
+#endif
--- a/src/transport/UTransport.cpp
+++ b/src/transport/UTransport.cpp
@@ -13,6 +13,7 @@
 
 #include "up-cpp/datamodel/validator/UMessage.h"
 #include "up-cpp/datamodel/validator/UUri.h"
+#include "up-cpp/utils/Expected.h"
 
 namespace uprotocol::transport {
 
@@ -67,9 +68,9 @@ UTransport::registerListener(const v1::UUri& sink_filter,
 	                                          std::move(source_filter));
 
 	if (status.code() == v1::UCode::OK) {
-		return utils::Expected<ListenHandle, v1::UStatus>(std::move(handle));
+		return std::move(handle);
 	} else {
-		return utils::Expected<ListenHandle, v1::UStatus>(std::move(status));
+		return uprotocol::utils::Unexpected(std::move(status));
 	}
 }
 


include/up-cpp/utils/Expected.h
```diff
@@ -22,77 +22,114 @@ namespace uprotocol::utils {
...
-	template <typename F>
-	constexpr auto or_else(F&& f) &;
-	template <typename F>
-	constexpr auto or_else(F&& f) const&;
-	template <typename F>
-	constexpr auto or_else(F&& f) &&;
-	template <typename F>
-	constexpr auto or_else(F&& f) const&&;
+public:
+	constexpr Expected() { new (&yay) T(); }
```
gregmedd, line 64: If T has any non-trivial destructor, this would be extremely dangerous as the destructor for the original `yay` never runs (its constructor _does_ run). Why aren't values being assigned or moved in the standard way? What is the intent behind using _placement new_ in this way? Isn't it typically intended only for construction into a pre-allocated buffer and not for reassigning values?

I cannot see a way that this _wouldn't_ result in memory leaks.

include/up-cpp/utils/Expected.h
```diff
@@ -22,77 +22,114 @@ namespace uprotocol::utils {
...
+			throw BadExpectedAccess(
+			    "Attempt to const dereference expected value when unexpected.");
+		return yay;
+	}
+
+	constexpr T* operator->() {
+		if (!ok)
+			throw BadExpectedAccess(
+			    "Attempt to dereference expected pointer when unexpected.");
+		return &**this;
```
gregmedd, line 123: Is there a more clear way to write this? Is the intent to call `operator*()` then take the address of the returned `yay`? Since it already checks `if (!ok)`, couldn't it just return the address of `yay`?

include/up-cpp/utils/Expected.h
```diff
@@ -22,77 +22,114 @@ namespace uprotocol::utils {
...
+	}
 
 private:
+	union {
+		T yay;
+		E nay;
+	};
```
gregmedd, line 130: Ah... I see why this is not likely to leak memory. A `union` doesn't allow non-trivial destructors, so it (probably) doesn't matter if `T`'s destructor runs. However, that does not meet our needs for `Expected`. We need to be able to return `T`'s with non-trivial destructors, like `std::expected` does.

You can see an example of this in `UTransport::registerListener()` - it is returning `Expected<ListenHandle, v1::UStatus>`.

As soon as the union is removed to allow types with non-trivial destructors, the memory leaks will appear:
image
(note: I removed the union and added a test case with a struct that contains a `unique_ptr` that is allocated in the struct's constructor so that the leak could be seen)

test/coverage/utils/ExpectedTest.cpp
```diff
@@ -32,7 +74,126 @@ class TestFixture : public testing::Test {
...
-TEST_F(TestFixture, SomeTestName) {}
+TEST_F(TestFixture, ExpectScalar) {
+	auto exp = parse_number("44");
+	EXPECT_EQ(true, bool(exp));
+	EXPECT_EQ(true, exp.has_value());
+	EXPECT_EQ(44, exp.value());
+	EXPECT_EQ(44, *exp);
+}
+
+TEST_F(TestFixture, ExpectComposite) {
```
gregmedd, line 85: Some non-trivial types should also be tested (e.g. `std::string` or a struct/class with a destructor).

test/coverage/utils/ExpectedTest.cpp
```diff
@@ -12,8 +12,50 @@
...
+	double retval = std::strtod(begin, &end);
+
+	if (begin == end)
+		return Unexpected(parse_error::invalid_input);
+	else if (std::isinf(retval))
+		return Unexpected(parse_error::overflow);
+
+	str.remove_prefix(end - begin);
+	return CompositeExpect{retval, -retval};
+}
```
gregmedd, line 57: Could `Expected` be tested more directly in the test cases? Most of the code here isn't really relevant to the operation of `Expected` and obscures what is actually being tested.

include/up-cpp/utils/Expected.h
gregmedd: Maybe a `std::variant<T, E>` would help. It already has most of the necessary logic and implementations around moving / constructing / copying values, and checking the type contained could take the place of `bool ok`.

gregmedd (changes requested):
Some changes are required. The current implementation does not allow for types with non-trivial destructors, but that is needed in several of the places `Expected` will be used.

include/up-cpp/utils/Expected.h
debruce: This supports the usage pattern seen in these tests: Exception_pointer_dref_when_unexpected and ExpectComposite.
I believe this came from the Andrei Alexandrescu youtube video. It is also on the cppreference page at the top of the observers section.
It must be the case that this could be written with a std::get<T>(storage) here instead of the redirect to the operator*(). Is that the ask?

test/coverage/utils/ExpectedTest.cpp
debruce: The function pattern started because it was in the example8 on the cppreference page for expected. I think it highlights the common use case, which is to check for failures in functions. The functions could be done away with, but that seems less precise to me. I would think, if anything, there should be additional test cases for assignment with expected, not lieu of the functions.

include/up-cpp/utils/Expected.h
```diff
@@ -22,87 +23,161 @@ namespace uprotocol::utils {
...
+		return std::holds_alternative<E>(storage);
+	}
+
+	constexpr explicit operator bool() const noexcept {
+		return std::holds_alternative<E>(storage);
+	}
+
+	template <class X>
+	constexpr E value_or(X&& v) const& noexcept {
+		return has_value() ? std::get<E>(storage) : v;
```
gregmedd, line 116: Based on the [cppreference page](https://en.cppreference.com/w/cpp/utility/expected/value_or) for this function, I'm pretty sure that the `v` needs to be forwarded:
```suggestion
		return has_value() ? std::get<E>(storage) : static_cast<E>(std::forward<X>(v));
```

include/up-cpp/utils/Expected.h
```diff
@@ -22,87 +23,161 @@ namespace uprotocol::utils {
...
+	}
+
+	constexpr const E& value() const& {
+		if (!has_value())
+			throw BadExpectedAccess(
+			    "Attempt to access value() when unexpected.");
+		return std::move(std::get<E>(storage));
```
gregmedd, line 123: This shouldn't move since it's the const version
```suggestion
		return std::get<E>(storage);
```

include/up-cpp/utils/Expected.h
```diff
@@ -22,87 +23,161 @@ namespace uprotocol::utils {
...
+	}
+
+	constexpr const U& error() const& {
+		if (has_value())
+			throw BadExpectedAccess(
+			    "Attempt to access error() when not unexpected.");
+		return std::move(std::get<Unexpected<U>>(storage).error());
```
gregmedd, line 137:
```suggestion
		return std::get<Unexpected<U>>(storage).error();
```

include/up-cpp/utils/Expected.h
```diff
@@ -22,87 +23,161 @@ namespace uprotocol::utils {
...
+			    "Attempt to const dereference expected value when unexpected.");
+		return std::get<E>(storage);
+	}
+
+	constexpr E* operator->() {
+		if (!has_value())
+			throw BadExpectedAccess(
+			    "Attempt to dereference expected pointer when unexpected.");
+		return &std::get<E>(storage);
+	}
```
gregmedd, line 167: I'm not sure we need to allow non-const access to the value when its not being moved out. We could probably get by with just the const `operator*`:
```suggestion
	constexpr const E& operator*() const {
		if (!has_value())
			throw BadExpectedAccess(
			    "Attempt to const dereference expected value when unexpected.");
		return std::get<E>(storage);
	}
```

test/coverage/utils/ExpectedTest.cpp
```diff
@@ -23,16 +29,361 @@ class TestFixture : public testing::Test {
...
+
+TEST_F(ExpectedTest, ExpectScalarScalar) {
+	using namespace std;
+
+	auto sample = get_rand();
+	auto lfn = [&]() -> Expected<int, int> { return sample; };
+	auto expected = lfn();
+	EXPECT_EQ(true, bool(expected));
+	EXPECT_EQ(true, expected.has_value());
```
gregmedd, line 55:
```suggestion
	EXPECT_TRUE(expected);
	EXPECT_TRUE(expected.has_value());
```

test/coverage/utils/ExpectedTest.cpp
```diff
@@ -23,16 +29,361 @@ class TestFixture : public testing::Test {
...
+	static std::uniform_int_distribution<int> dist(0, 1 << 30);
+	return dist(mt);
+}
+
+TEST_F(ExpectedTest, ExpectScalarScalar) {
+	using namespace std;
+
+	auto sample = get_rand();
+	auto lfn = [&]() -> Expected<int, int> { return sample; };
+	auto expected = lfn();
```
gregmedd, line 53: Is this doing something specific? Wouldn't regular Expected construction be adequate? What does this test that isn't covered with:
```suggestion
	Expected<int, int> expected(sample);
```

test/coverage/utils/ExpectedTest.cpp
```diff
@@ -23,16 +29,361 @@ class TestFixture : public testing::Test {
...
-TEST_F(TestFixture, SomeTestName) {}
+uint64_t get_rand() {
+	static std::random_device rd;
+	static std::mt19937 mt(rd());
+	static std::uniform_int_distribution<int> dist(0, 1 << 30);
+	return dist(mt);
+}
+
+TEST_F(ExpectedTest, ExpectScalarScalar) {
+	using namespace std;
```
gregmedd, line 49: Minor: Is this used here?

test/coverage/utils/ExpectedTest.cpp
```diff
@@ -23,16 +29,361 @@ class TestFixture : public testing::Test {
...
+		EXPECT_EQ(1, PairDestruct::cd_count);
+		EXPECT_EQ(false, bool(expected));
+		EXPECT_EQ(false, expected.has_value());
+		EXPECT_EQ(x, expected.error().x);
+		EXPECT_EQ(y, expected.error().y);
+	}
+	EXPECT_EQ(0, PairDestruct::cd_count);
+}
+
+TEST_F(ExpectedTest, Exception_value_checked_when_is_error) {
```
gregmedd, line 257: These identifiers must not contain underscores, according to the [gtest documentation](http://google.github.io/googletest/reference/testing.html).

gregmedd (changes requested):

gregmedd:
Do not close pull requests with unresolved feedback unless the code is being abandoned. If you need help fixing the branch for this PR, let me know and I can help you with that.

.github/workflows/ci.yml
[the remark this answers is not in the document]
debruce: I created the new PR to resolve this comment, which is marked as outdated. Greg wants it back here. I guess I understand it didn't need to move in the first place. Sorry about that.

test/coverage/utils/ExpectedTest.cpp
debruce: This test proves that the implementation can handle expected and unexpected types that are identical. The motivation for having to tag the unexpected type is driven by the need to deal with this possibility.

I think the std::expected thing is interesting, but would be better if the unexpected type where maintained to be something derived from an exception, if the calling code wanted to easily default back to exception behaviors/patterns. Having done all this work, I don't see this as being worth the trouble. It is a dressed up variant, and more code that could break.

include/up-cpp/utils/Expected.h
debruce: This will be fixed in the next push.

include/up-cpp/utils/Expected.h
debruce: I added a test case for value_or. I tried the requested change but it does not compile. Removing the static_cast allows it to compile. I think there are other casts where static_cast could relax the return object initialization a bit, but it would require some focused attention. Not sure if it should be pursued now. I would recommend it should not.

include/up-cpp/utils/Expected.h
```diff
@@ -27,82 +28,137 @@ static_assert(!__has_cpp_attribute(__cpp_lib_expected),
...
+	BadExpectedAccess(Args&&... args)
+	    : std::runtime_error(std::forward<Args>(args)...) {}
+};
 
+/// @brief Required tagging type for cases where expected and unexpected type
+/// are identical.
 template <typename E>
-struct BadExpectedAccess : public BadExpectedAccess<void> {};
+class Unexpected {
+public:
```
gregmedd, line 43: Redundant access specifier
```suggestion
struct Unexpected {
```

include/up-cpp/utils/Expected.h
```diff
@@ -22,87 +23,161 @@ namespace uprotocol::utils {
...
+
+	template <class X = E>
+	constexpr explicit Expected(X&& rhs) {
+		std::get<E>(storage) = std::forward<X>(rhs);
+	}
+
+	template <class X = U>
+	constexpr explicit Expected(Unexpected<X>&& rhs) {
+		std::get<E>(storage) = rhs.error();
+	}
```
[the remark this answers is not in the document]
gregmedd: @debruce - You mentioned that this did not work. I tried it and it does work. However, there is a bug in your gtest code. You are using `auto sample = get_rand()` where the signature is `uint64_t get_rand();`, then trying to implicitly return `sample` as an `Expected<int, int>`. Once the test code is updated to make `sample` an `int`, the tests work.

[Compiler explorer](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGEgKykrgAyeAyYAHI%2BAEaYxCAAHACcpAAOqAqETgwe3r4BaRlZAqHhUSyx8cm2mPaOAkIETMQEuT5%2BXIF2mA7Zjc0EpZExcYkpCk0tbfmdtpODYcMVo8kAlLaoXsTI7BwA9HsA1ELKACIAGgC0AGJ0mB6pAJ7EeMAIBAAqmKoEIIdmGjMkkOHkEL2iXiIxAUhyIsIQmEO2GQtDwqSUh2umwY6CY9QYJg0AEEDoSSUchJhEQQEYcIgB5D4ASTc2EONHoEAUq0O%2BAm4MhmHQhwA7oQEPC8DCRSQANbskiHJjofD4sRkg6HML8YgsPHZQ7ETDAZr4IyHNBPF5vAiHVAi8LQhBogB0Gr27sOH2dMNSxFQwGIrCVOPhiKYyDQLFShkeYWAhz1BDieDEMOaiL1WCVADcmHQmNF6IcvDi4mHPcndTDUFQw4ciTHkLTQjsGBiAGpxTICf4ujSi53NrXpvMFovhgie94EdEgA4ixcuphNhEukjAPaottKBR7YIs7ARITYS5mfvuz0nC6XVuMJSXJlYQR4GijBurzBni/EsnmADMr5lnWcjKAA%2Bm4yjgXIzLBEIYHYOcyjYG4HzYKcYEABJ/mY/5YDQ4SHKBEFQWBMFMnBCFIShaEYdhv6/rhYQol42YmP%2BbgTFgqg7KkU7/tgOGAQwLFsRxBCPKkmBgQQQaEAo7GCYxwmiYi7FuHmLyGPxSlEmSzBsAoTaIl4fqoEQaC0CAICQnQMImAA7FYDHkg2hxKC%2B4S0IcspUqk8ainiw46oOjCioiChigQw5wm41jWLhIbCiaYRKpGmAKDCRCVrSRpiO5BDoNZ3xSQ4Qouoc9JbIcqIsIQQoWqgTzuQgmy0MKsSGpgqQGDswp4CwbD4HitSPG6xITPqyBgUwmVxAQEBgGAYFgQgs1gcgqSpDNBCyXgELJhAK2bdtqLRGBJU9Mm6DrGShz3Q9j2PeYZgAErdb1JlmRZnjWbZtAKNZ2CqKV12iuKBVFSAl1legL2rOxzl6cSBxHAAAgZiJfNGJDNI87leNEEyEIKCrEJDxUg1dQruujRp6sQsowpS1K5ZgVBxIwOxkxT0NU7DIC009T0znOByMC6J1GhzRoiZgkuoCwewinsJ17P9hCPHsMPXULwsPREqDslsNLlugqDID4jBNPiI6HGZOZ4Fg/UMJKMIcvLQto45SP8l4DiHBYyrA6DQpEulmWHH8qSE9uvPEKWjhsBdxD%2BuTPt/hoSRB%2BgIfU%2Bg4c7JlEA8n8XHWQnL7J3EacQC9wCMCmyCB8H/PXYXGUKWYZg8hnDmnJnyNJMm0YGMmhzqRJUmYy6s8NsQwAKQJmfZ63ocFxHCgQESC9dwAbOYe%2BzxVzSLwjDFJI9ZeFRXicDdJNckFyN8gDqIqmupO%2BL4pECn9yx%2B9yco5Aev5%2B6I0HqjQ4aNogvHZocd6ABHLweAjTCiaMAYAAUp6IhCqIXcoUjSHB1g1QwwpSzELQZJTAQsMxamfI4UQtBxpEhHj1EaE9xJUMxkiRSZIUSzRhHIcIbcGoZ2JDHIseBkCCwvmgdsyYQbkyERQiAciJhEWEevQ%2Bvd/ynF5OzJgXhaD8SRoSJIaiFF%2Bg0So5RIi4ZmAPg4nRej8KGOMeAlyZiLElXJiVbchBrF2NUQIdRgkHGGgQNyKOBVcYNwgMQSJgCrBgNkSEyxviQb%2BNtLY9eEAwmOL3hEqJ19YmYGflDFgqAcxlISdyJJwDB5mNYWPNSHF%2BGZWPvPb%2By9UnyJ8UQzJUiAk5PzuU6yYQwJsJ2DJUgXT96H06X/c%2BQ8r4xKDHE8ur8SDv2IHDDiX8l7YF/rvVYACJ5AJScjIe3jFGNXkTw8Jj9iAlzuRMQ%2BhwGCoG%2BLxW0PsuoEC2K7CYpTEbnJAVcrxaT%2BkWIeQUohqcn48gsYfd5nzvndV%2BU5f5gK1lMAbqChpnis43Ksfk95TyXkovCWiniGLzkWGxcQIFL9KnVOfqUhG/4GWEquX6PAeZkwyKHmyYF6zqFcr/GAiVLlIHQNgXWIkBUXhbSFJcc2DpDjVOhAaWsvMKHsn9CwEE8VLC4WYc09hk8uGsCxrM7B3DBI9PDgYSOedYb0rJBI7cQqAKvkOBoFeJLyZuuuhAD4VLCm1NLrijZLKqk1MSfUy5kK%2Bm3JDUKCAIzYbqUdWS8JUbomirxWUzZrKE11I9cm3pEx%2BnpvQME%2B5WbdYcVze8gtYih5mM2Q3fibgm00xbbwgS7KxWnKLQ3DhejS3xviYkjxnaq1DwtePdS7SYTnEnV6Xh1b0kDJ6kM20daIDnAjUU%2BpF8u0vx7epcNw7x2YGcbzN%2BH8OInuHVG%2BdZieUr2Xa0twa7DgbvYno1tO7%2Bl%2BIPUiIJ/a9luDfXmyNiSPUXqzt2zAvbc0juLY%2B2pLoKWctMVnb9YG01BJhXWttSGSlitnQoF096k3goArUJQd0Hq/qtdPG1nSDnbsVQ9INUHcm8YcQsueSzC1QmLWMrZxAdmwd48OpZZy%2B5Mdwq4V8jTiVQtudEVAnhDhrQUGBPM3gylIrSR8r5tK%2BLIc7VnI0AKmW81au1YzYgqzMEcNUm9P8GOfqI4ulNNbbkQeQAEpqcQ8SKj054F5MKaU/Ls1%2Bhz6GcWbNc%2Bgdzxi4hef5eK1CfmpMNwIyvYjQ9NS/o4f%2Bl166%2BNmM1IJj4mqxBeGkk/E9onwk5gs/I1F1mksdoa0cMxjmcVGZM618zHDri82vRxW9RyGNRxXpqSaDCNqzQw0VqGz7dnqXgxAHrpWL6anK00zAo9LVtNq4B%2Br2nU1WOa6ZtrYEOunp668qc1KBt0qG0R1LTnXYTZe9N9is20O9sW1hkrkndvbJfXBn%2BPWAtJHOw9kLViYXhu61N%2BLaT3n/bR1nP1i0wAg7x6sZZKXifDwQP6EULdc52I7sXMrAPL4PReiw1hto4QRiLjCUHLyRQIlduQuxbpu4nfs0kMbznIe%2BbvcVh9qP0fmJ009lrZnKVdcKUTsxpOloU511T9ntOaQM6Z3W1nW9zdmOet3Hnl3bP883trtrIuxclk0fnKXPc1eA/S3GtliuFtFY5TLr9QWMe7vI%2BShFzzetvPCQbkndYIAm892blDFv6f2mtyzzeddc8O65073aLu%2BdGwF53eFtceSi7Cp820Ev17%2B6jxz%2BXzKobzb7b77Ng7l7K45XhxPJc1cx41494N9fEUTz18ljnpOs/mepxzpp%2BfGc5xt8X%2B3WdHdmGd9GavaVBdz6T6FV2LefcUI74HuXaWFch5LVe9D6kYM5qHUtlXVOx8N8nyYxIyx0sxx0KUiyDChAACp8d7k08kgjdyd1phcc9ZdN8rcd8i9BcS80CD9y8j9K8T9YQjYsBpZOY5Y9184PdEQm9xcB9rp79pVZdu85t39w8R9R1ACtNp9MdyZscoC7QpJICSBLgf5k9MVCNadEDV8S519ad0CC9MD15bccCacy97pudCDXcSDOYZYuZER9V0gwgqwr9b9Jd4YH8WDD4w9CsODsMuCXI%2BUBV2AV5NlNJUxBAb1ZlP8h9c0Y1xVCMh51spEZo5oWgydNkpQTNUAnYTMldsBSAV59Z9YXoAB1REc2BgMADgW0MQWgAvR1AeXRHMGI4UCAUsVEXyPVOxBGaXALIIm2EIgReaVQvA%2B6SI4zLAP2XoCcOI9gtkU9JaDomaVOJgR4Po2wgYvXJI/WIYl%2BKIsg2WaZHMeIxIi%2BQ/OtLqJBFBDKJEYgxMKkPnWkOKSwawQ4bIro2Sf2JoCcbIrY5Bema2CwpghpF4lGTUb2ILfue6TUTGIyCML6f0H6KyGyRwAGRpdTHEP1Q4TUYiSCaCWCeCRCZCVCdCLCCBCkM4K4W4egB4Z4V4d4L4H4P4AEIEEEWQAUKELKI2U2JEFENEDELEUsXEfES8d444KkesBkZkVkdkO4LkHkPkPaA6BqKKCUGkKUUUOUHmZUVUbIdUdk7UEgJMA0I0E0XZAKS0fEm0O0B0bsZ0VIZhUkdk70SUsyQMYMUhesAXRWGMBgOMc0JMFMNMJUQhLMcMMcAwCcH3UgisdkqsFgGsOsWkxsCMFsKRe8RELsbVXsc8AcUXKRCUSUpgT0wsYsPEacXaMWZWJcFcMM%2BWDcLcCM9sDKfcQ8Y8U8OMtk1ya8K4O8Esx8ehV8PAd8UM5sL8KszxJiESbwMSNwYAZMCYPYAcjKAgF0BAe7PSbs1SarIMHERWScv44yJfIkLwTIc0UyIEi2X6UEuyayJQ/OW3edNcgKTc8ybckE/6QGEACjJgk8jc76C8v6ME68nwwIgDOtL4dRaOWOKRWEUc%2BMayL8iQz1IE/OH1LOTUV6UsO0CgoQr0bAIQD4MCa4ZhYbIiJQNBI2JQVvVIC0egQwIhBgR2f0BgNgQQX0csQcscleEop2Dk%2BQVIF5eNVOJ2NSC5IApIOi4UL4ZoU4e0BgZirVF4NiDi7gqCmCgQbmeC74HoWyXsXVWk6ipULabcfUAQNCyCo4RQBqEKHC0yO0OsPI/8iYOjQ4TCWaM/OvOECUmEMISaOWTSpIT80cylXRfRKgNxExFeAAPxcomDcpcQMSMW8qJSSAktdiksRBktUDkttkUtpGUpXH3VEHxCcthMwtgtoHxhCmAHyOiHyn0rws2AIEyGzEUtHLo1cMaObm4oYrkFSGAqEGQUOkY2qqmk1VKK9EwD4oEqaparBzEtASKMCOQUED3kkBkkOB7TAjnPrXPQaI6s2TmsVjAiwEdm5n2zeKSGCOblLQIC4CSCSH/AckTAWl2VkNR12t5lLFfGVLAmMLWqlGFPkoJA4mMMUl5GeogA0FmS4Gq3UkOH/A0E70fyBy%2BoCpYAICj0XTQiQpQryTsWAtmRDREDEGaDRoMGIAWsN0EHclYB6laT0Rmrmon22sMThFoCoFdmAwnn8CsFE38FOBeVEIEiE3znUmMNmQ%2BrZr%2BRYIUAJvoAJSlUkIpqNn1VpqpsEtBuRJojAg%2BFejkCOVi1oAgAoTN22tltQnlsVqOTvxkI1skK1uQuwAAEUuRBbMBZk78UCZbqJtazaLbR4rbDgoD1bP1YbELkLrhEb15kbAlQZMaMbGFmgcaSc8aBbnbN0SbSEybRbIQjYpbN0TB6aUUmaWbPrby3AuatRPDeasUWCYMnbCaCMwVUcxbKD3VJbqa46V5jaUKiQ4Jlb9NVb1a7aUTvbG6Tw1bJcDb265bHbI7CbrbJd8MYagCiQ4bvbfb85/bUaQ7sal8ECI7Lbo70NZrY7QaK6k7aaU6GaD506eRWa2Qs6c7Nl%2BR4xPq%2Ban8gVLbhaRqV4K6Jb3Kpba6L566FalaIAVae715Da677bkLP69be7kDKd%2B6Hbzah76AR729bbUd67B7LbZk3aaiPaJ6p6Eb/KCBvD6Cg7F6idjD8ao7aaY6cQ36h5t7qbk7U7Gbmaj7M67F1Jz69ojBua86BigFc9C76CM1oHVdtrXj464Rn69FX6IHO6m7v6W7f785/737AGG6pH9awHTcJGEIoHkHK6GCx70HB5MGfbsHcGdYOwpt6QCGuGh4iH%2BG16CAN7yGt6E6apqHd7aGD76HDhj72bB9s7BBZkWGXgjAr7LGaceGVEAn4w65u4ERaB8j4ZQahHH6nHRHnHpaEHFHrgu7m64s270mO6lHu6VHjN4HNbFGkHnbYG/dQc3tnl%2BH5G9JF1ujbRlB8x04Qnw7bRVBUciHHh6izEWmUEIAiHVB2HbRHho1VAIBVB1hDhHgIBxnK0mMRb9GvasGkbRyUa24hE8AkEH0l6K7VBbH7H5ry6nH8ZSH17SbO9KGnGd73K9606PGvGs7NlbrdnJlZJ1IBniBFJ/GX4L6gn87JDRsb7eY9RfIwI3m2svnWmf4RnZmEnlmL4n67FN1xG8m5bgHpGcmaj1GsWinJs1HTm4Q8Labp02V1aXQUCMXIGpnZlUhj7pmaWTbzbHh6Xj7xm9GXIDGZ7YZ/am1tndmw6kgDmjmrniWjZzn3KyGTm3ibnKaXH7m3GU6nnGGtF3q/GbqGAdnXs%2BIfmOJvnv8l6QXwai7S0mAIWoWCtDXh14XxnEWH7kXknUXq60nSn8nMmpGf7cn3W5bPXCnQHinwGJX7ZN1yWylKX8NmWNG6X7ZGX1HHa2W43FJOWXjPb4bDH1mJhNnA61pUFhXRWLm7HxXyazmxXN6%2Bms4qGaalX96VWM62aXn/m82hQPn9XIJYWBI/moYAXgBgngWg9n8KkLXpIFAW3YMbWjk7WHWQ2Un0XfXtasXvXcXo38XA3CXs8E3zb4WbbKd43o3E3KnYYqW92U2t3Y2IBUG/792F2WX5nZlL31aOXx6VmM3eXrp%2BW8Hx2C2nHDmi3jmKGzEK6pXibLmK25WgPbnFW9EHm6GG2T6mGNWcHeYx2Mx0A22YWUEjX4CwmglzWIXUPUFMP9Wp3Zl7WuDS2RGXWX6a71H/XsnW6V3b2CmQH28%2B6D3t2j2dHx9Vgb2jaynWWuPyp8Nn2uWrkeWjHvHhAriHAf24Q/3pWwOHGQ2QPpqlPZXAjIOFWa2YPlXD7PG1WOaDXWnu3rJe3%2B3GVXZvnY3U3uUkX5XxbqOxHaPV3dbsXGO/68W3OCWSn%2BP8nHad3JcUCXQmXmPD3tHypgvbOAH/POOIv0A%2BOYuB7BP4vRO02MHVnM2/aNmA6rpGgE5ZP9nf3y3lPKPJWSuNPuDq2aG639PnnEPfHkPJ36VLPcvYYIBrPp3S7EmnWqP140WXPmP6P3PZHYZ6mzF67hufPg2wu4u798MQvz2k35ueOXRovhrB4mnDhvnThRyCuQKL5hnum8bentrrqiHkB0O0BE5K2kgdu9vrihm8b4WemJmpmZm5mFm/lLuNpsRvsGarAuUy6wr7ummG11FQeZPvtClT43vT5FuyPjlgA1ukkLQru/uTVAe7OJ6zEfLIf9vKUsUfvrvBBLhWbsfJVHWrkiH8frjrJie/vN0A05WJOs3kPUaofdumnhXaeHB6f0fE4mfJ8B2RXiv/2S2RfgOKvAPc9qvXHavVXG2GvefkOInAXOGReTWcUVebOEnAsqeacUX%2BvXWZfadEHzauBZkVf%2BffvE49fnLFGl2ZGfWReP7vP13fPzdzfY3d3TdFuH9vflugvKcUeA%2BBOfe7FEvc9A%2BhOEuz2HDhVw/frtvWmueoebeSfoaxOyRWfsvs22vpP9u0/9uefU%2BHu%2BeUABe8babmfgX2nadC3FPi3wPJey3xeW/uCtPE7oO6aFf4OpOB1Gurey%2Bmn%2B398wacUi6deuuH8evZcjeqCTf7fvfLeU%2BUFi%2B6fK/bfBBl%2BMmsmRuXeve9/lGPeZvXfw/Av28FvQvz/Yv734v/9EVQ/BGq0He7/k/ret/M%2BX3uXMv32hQ56NmfEKYzMxuAEQyAXyOgFSJi4mQCgbAInjk7d8dOvfR5v31Pqas1efbIFublw65JMBUTMwDEzibS4Z2EHLOB/UwivR6QqRVopzg0L191CsIYgOcwYEc5nogjNgQ9AX5V0aObrTXm0SRDH9u6y7TzrPwEGTd9%2B03IlhwNpyPQK6oOTdL72zwP5no3xVKsOHB62gDysMI8o8mmbGtOB90eukhVeiO1NCvOfYrXkjgoFTCbeP3C9DWK4E6ByRYWN8BdCi48Ql1GQYwMtz2gVBXOV/qoNOCODGB2g9uJvB/7ic/%2BknYGD8myDwC04YAnoJAOgGMBDYBABISQEQGpMauqAhhkr3VZD8UOrDLARr1a7%2BB76t3CgVQJoEzF6BIvR6LJBYH8DnBYg5wdwLBhL82hj0N3l/REFyN7ejA3oaxz9zsdvBAgp%2BqnEUGj0eO/g%2BgXonUEShNBheZQpvHJT6D4CQwxRiYLMEV4LBbuc/PhlMI347Bx7BwePxcHJE3BHghaONw3wCDfBIoOYcDxWQBDghdQlYYeQiHZ9iQufWejl1iEYpsgu3aWCALaypCGAMGbIXc105998hCHQoTCm%2BamcQA5nbAdw1BZmt/mJQggUQNQDxMKOmncgY70oHUDaBqghoexmYEGDZBXOcYe0OdbG9eBpvGnD0KEEMdRu10O4fIXEHsjOR5UMYS0MegwpG%2BeiK9kZ0pHzCLQQUJYTCjCFhw1heg4VqyIejGCFaZg0vAIMPzH5tC%2BiRYvoXi7UFbBvDdAP7hCH3DnBj0a4WtFuGDCHhW%2BZ4cRlaH9xzRzg%2BURvEFyRCc%2B0QtnpsziECAQR7MZQLJAhFQiiu2nXIXB3hED8J2JnYoYE1KE0jcBoyfAS9DxEEiX%2BBvBRvkw%2BCkjah6xAIZKKYHNCLhdIoUVwMZGL9mRdo5wRIK9bO8mO5YowXyKkGbtuhFY%2BTtMPXiMtHRagmUYcGWHujdBhSb4MqMMGCD8mOw82qWOFjmCq8%2BxUguzHILcxDCMRQQOWFoJmF285wzUZaIejWjPB3IlUfdEeG9jOKQQ10Y9CHHfD0ujSb4jCSOBLkASEJQCFQHwiVQYIYEekNcB1rYBdIZIIhnqDCBPdciC8ZALMmbDNABCp8FYqnSZrCtrI1FQCiACZDasCAAAcX0x5VMAwFKJnvFPjgTXSwAFHNtRYKK0IgYERusEHlpe0hAgHILJCXwBUAOA6wKyBwH8C8A/A3AXgKgE4AnEAe7kTYNsDUi4QeApAX4BwC0BU5SAsoEAP4F%2BpsTJAnEzQDxM4C8BAYv1CSVJNIBwBYAMARACgFtJ3AyAFAYJNGGMkgBgAXAXCHwDoBVhAY39FSaQGiBhA8YnAMSS5OYDMD6Q0QbQFdHcm8Aow5FAgPSAYDZUnJWACEMADcB5FAY3E0gFgD1BGBxAkk3gPgCNC9BqkcUrQEEFiqWxBUqU8gOuLYk5SzoQYZgR4CwBOS9oLAAKaQC1R6YlAu3JKXlUAqpT1gVAAwIvA7AtkRQ9IaeHVP4CCBMa7AKQDIEEA6U1ATk3QJbwMBGAUAmPfQPtEBiwBMYlkrCfQGor1TRgVxBgLKGklNR8QcUsnlxGAymBTilgLgKdUuC0ARyEwMCEBIYA8ThKbFFaSXBqB1BsgLgHENMA6BBAcQQwcoJUD0DpBMg%2BIX6SDKKD4hAZIweIJb26C9AGg8wCGfDNqBXQ%2Bg8wGGcsDhlzABgKM3GS0CxnAyuA6wBQEJJ2ASBWJnADiaQC4k5TeJHAQ4KoASB7xLgE1aapGEODWSXQ/4AcbgEICKgAIJM3gFpJYnrBZJ8k/QJwCUmkBapksumapI4DqSQAmklSesF0kGSSqMcHBqZKjD%2BR6AxACIDak4DMzWZ7M4AJzKkAugxJQoAWSJT0BDThAjCUadIEdmTT1AhUmaaQBFBBhUgAUqmexOUmFSGZVUWcJCEMpMyWZbM4EBbObjczeZEADwOZINkL5/wws8SWrPWAIhlQowd6YpN4ByzfqCs0gAzOVmqz2p4suSQpM4D/gg59MtSRnIrlSyOAZgOuYrJFmZztpMZPwJICAA%3D%3D) - note that `auto` was changed to `int` on lines 218 and 237.

include/up-cpp/utils/Expected.h
```diff
@@ -22,87 +23,161 @@ namespace uprotocol::utils {
...
+
+	constexpr Expected(const Expected& rhs) { storage = rhs.storage; }
```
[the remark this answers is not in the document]
gregmedd: If [this change](https://github.com/eclipse-uprotocol/up-cpp/pull/124#discussion_r1628657040) is applied, then this comment will be obsolete. Otherwise the problem still exists.

include/up-cpp/utils/Expected.h
gregmedd: [It appears to compile for me](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGEgOykrgAyeAyYAHI%2BAEaYxBIaXKQADqgKhE4MHt6%2BASlpGQKh4VEssfFcibaY9o4CQgRMxATZPn5cgXaYDpkNTQTFkTFxCUkKjc2tuR22EwNhQ2UjlVwAlLaoXsTI7BwA9HsA1ELKACIAGgC0AGJ0mB7JAJ7EeMAIBAAqmKoEIIdmGjMkkOHkEL2iXiIxAUhyIsIQmEO2GQtDwySUh2umwY6CYdQYJg0AEEDoSSUchJhEQQEYcIgB5D4ASTc2EONHoEAUq0O%2BHG4MhmHQhwA7oQEPC8DCRSQANbskiHJjofD4sRkg6HML8YgsPGZQ7ETDAJr4IyHNBPF5vAiHVAi8LQhBogB0Gr27sOH2dMOSxFQwGIrCVOPhiKYyDQLGShkeYWAhz1BDieDEMKaiL1WCVADcmHQmNF6IcvDi4mHPcndTDUFQw4ciTHkLTQjsGBiAGpxdICf4ujSi53NrXpvMFovhgie94EdEgA4ixcuphNhEukjAPaottKBR7YIs7ARITYS5mfvuz0nC6XVuMJSXJlYQR4GgjBurzBni/EsnmADMr5lnWcjKAA%2Bm4yjgXIzLBEIYHYOcyjYG4HzYKcYEABJ/mY/5YDQ4SHKBEFQWBMFMnBCFIShaEYdhv6/rhYQol42YmP%2BbjjFgqg7MkU7/tgOGAQwLFsRxBCPMkmBgQQQaEAo7GCYxwmiYi7FuHmLyGPxSlEmSzBsAoTaIl4fqoEQaC0CAICQnQMImP4VgMeSDaHEoL7hLQhyylSyTxqKeLDjqg6MKKiIKGKBDDnCbjWNYuEhsKJphEqkaYAoMJEJWtJGmIbkEOg1nfFJDhCi6hz0lshyoiwhBChaqBPG5CCbLQwqxIamDJAYOzCngLBsPgeI1I8brEuM%2BrIGBTAZXEBAQGAYBgWBCAzWByDJMk00ELJeAQsmEDLRtW2otEYHFd0yboOsZKHHd90PQ95hmAASl1PUmWZFmeNZtm0Ao1nYKoJVXaK4r5YVIAXaV6DPas7FOXpxIHEcAACBmIl80YkE0jxuV40TjIQgoKsQENFcDl1Cu6aNGnqxCyjClLUjlmBUHEjA7KT5NQ5TMMgDTj2PTOc4HIwLrHUa7NGiJmAS6gLB7CKezHXsf2EI8ezQ1dgtC/dESoOyWw0uW6CoMgPiMI0%2BIjocZk5ngWB9QwkowhycuC6jDmI/yXgOIcFjKkDINCkSaUZYcfzJAT2488QpaOGw53EP6ZPe3%2BGgAJyB%2BgwdU%2BgYc7BlEA8n8XHWfHL5J3EqcQM9wCMCmyAB0HfNXYX6UKWYZg8un/inBnSOZ8m0YGMmhzqRJUkYy6s8NsQwAKQJGfZ63IcF%2BHCgQESC9dwAbOYe%2Bz%2BVTSL/DDGZw9ZcFRXCf9dJNckFyN8gDqIqmupO%2BL4pECn9yx%2B90cg5Aev5%2B4I0HijQ4qNogvDZocN6ABHLweAjTCkaMAYA/kp6ImCqIXcIUjSHG1vVQwwpSzELQZJTAgsMxamfI4UQtAxpEhHt1YaE9xJUIxkiRSZIUQzRhHIcIbd6rp2JNHIseBkACwvmgdsyZgZkyERQiAcjxhEWEevQ%2Bvd/ynF5GzJgXhaD8URoSTOaiFF%2Bg0So5RIjYZmAPg4nRej8KGOMeA5yZiLHFTJsVbchBrF2NUQIdRgkHGGgQNySO%2BUcYNwgMQSJgCrBgNkSEyxvjgb%2BNtLY9eEAwmOL3hEqJ19YmYGfpDFgqAcxlISdyJJwDB5mNYWPNSHF%2BEZWPvPb%2By9UnyJ8UQzJUiAk5PzuU6yYQwJsJ2DJUgXT96H06X/c%2BQ8r4xKDHE8ur8SDv2ILDDiX8l7YF/rvVYACJ5AJSUjIe3jFENXkTw8Jj9iAlzueMQ%2BhwGCoG%2BLxW03tOoEC2C7cYpSEbnJAVcrxaT%2BkWIeQUohKcn48gsYfd5nzvldV%2BY5f5gK1lMAbqChpnis43Ksfk95TyXkovCWiniGLzkWGxcQIFL9KnVOfqU%2BG/4GWEquX6PAeZkwyKHmyYF6zqFcr/GAiVzlIHQNgXWIk%2BUXibSFJcM2DpDjVOhAaWsPMKHsn9CwEEcVLC4WYc09hk8uGsExrM7B3DBI9LDgYCOecYb0rJBI7cQqAKvkOBoFeJKyZuquhAD4VLCm1NLrijZLKqk1MSfUy5kK%2Bm3JDUKCAIyYbqUdWS8JUbomirxWUzZrKE11I9cm3p4x%2BnpvQME%2B5WadYcVze8gtYih5mM2Q3fibgm3UxbbwgS7KxWnKLQ3DhejS3xviYkjxnaq1DwtePdS7SYTnEnV6Xh1b0kDO6kM20daIDnAjUU%2BpF8u0vx7epcNw7x2YGcTzN%2BH8OInuHVG%2BdZieUr2Xa0twa7DgbvYno1tO7%2Bl%2BIPUiIJ/a9luDfXmyNiSPUXqzt2zAvbc0juLY%2B2pLoKWctMVnb9YG01BJhXWttSGSlitnQoF096k3goAjUJQt17q/qtdPG1nSDnbsVfdINUHcm8YcQsueSzC1QmLWMrZxAdmwd48OpZZy%2B5Mdwq4V8jTiVQtudEVAnhDirQUGBPM3gylIrSR8r5tK%2BLIc7VnI0AKmU8xam1YzYgqzMEcNUm9P8GOfqI4ulNNbbkQeQAExqcQ8SKj054F5MKaU/Ls1%2Bhz6GcWbNc%2Bgdzxi4hef5eK1CfmpMNwIyvYjQ9NS/o4f%2Bl166%2BNmM1IJj4mqxBeGkk/E9onwk5gs/I1F1mksdoa0cMxjmcVGZM618zHDrg82vRxW9RyGORxXpqCaDD1ozQw0VyGz7dnqXgxAHrpWL6anK00zAo9LVtNq4B%2Br2nU1WOa6ZtrYEOunp668qc1KBt0qG0R1LTmXYTZe9N9is20O9sW1hkrknJqbbeQtnb1k9uwcO8dgLmdzsPZC1YmF4butTfi2k95/2sdZz9QtMAIOierGWSl8nw8ED%2BhFC3XOdiO7FzKwDy%2B91nosNYbaOEEYi4wlBy8kUCIXbkLsW6buJ37OZzG85yHvm73FYfZj7H5idNPZa2ZylXXClk7MZTxaNODd0%2B54zmkLO2d1s51va3ZinrdwF5d2zwvN767axLqXJZNH5zlz3LXgP0txrZarpH6uOUK6/UFnHu7yPkoRc83rbzwkm4p3WCAFvfdW5Qzb5n9p7cc83nXQvLu%2Bdu52h7oXhsRed3hbXHkkvQqfNtDL9ewe488%2BV8yyG82%2B2B%2BzYO5eMfR14dTyXLXCedePeDc3xFE8jfJZ55TvP5n6c86acX1nOcHfl%2Bd1nV3Zh3fRnr6lUXS%2B08hRdh3gPFCe%2Bh6V2llXEeS1XvQ%2BpGDOah1LY1zpynxb1nyYxIzx0swJ0KUiyDChAACpid7ks9M4zdqc1pxcC9Fdd87cD8y9RcK8sCT9q8z9a8L9YRDYsApYOZZY9184fdEQ29pcR8rpn9pVFd%2B85tv9o8ACOVQCtN59ccyZ8c4C7QpJYCSBLgf509MVCNGdUDN8S5t9GdsCS9cD15HcCCGcq87p%2BdSDPcKCOZpZOZER9VUgwgqw79H9Zc4YX8ODD4o9CsJ9sM%2BDnI%2BUBV2AV5NlNJUxBAb1Zlf8x9c0Y1xVCMh51spFppZpmgqdNkpQTNUBHYTM1dsBSAV49Y9ZnoAB1REM2BgMADgW0MQWgEvR1AeXRHMBI4UCAUsVEHyPVOxeGeXALMI62CIgROaTQogu6WI4zLAX2HoCcJI7glfOFRaHo6aFOJgR4IYxwtkCNNIvWMYl%2BOIqgmWaZHMZI1Ii%2BU/OtTqJBFBdKJEcgxMKkIXWkWKSwawQ4fIvo2SP2RoCcfIvY5BOmK2GwtghpD45GTUL2ILfuO6TUDGIyCMT6f0b6KyGyRwf6RpdTHEP1Q4TUYiSCaCWCeCRCZCVCdCLCCBCkM4K4W4egB4Z4V4d4L4H4P4AEIEEEWQAUKETKQ2E2JEFENEDELEUsXEfES8b444KkesBkZkVkdkO4LkHkPkXafaeqSKCUGkKUUUOUbmZUVUTIdUbk7UEgJMA0I0E0XZfyS0Ykm0O0B0bsZ0ZIZhUkbk70WUsyQMYMUhesEXBWGMBgOMc0JMFMNMJUQhLMcMMcAwCcAPSgisbkqsFgGsOsRkxsCMFsKRe8RELsbVXsc8AcSXKRCUWUpgX0wsYsPEacHaUWJWJcFcKMuWDcLcGM9sdKfcQ8Y8U8JMrklya8K4O8Csx8ehV8PAd8SM5sL8OszxJiESbwMSNwYAZMcYPYEc9KAgF0BAe7PSfs1SarIMHEBWWcoE4yNfIkLwdIc0UyME82H6SEuyayNQ/OR3edLc/yXc8yfciEv6AGEACjNgi8ncr6G836KE%2B8gI0IgDOtL4dRKOGOKRWESc%2BMayP8mQz1ME/OH1LOTUF6UsO0GgsQr0bAIQD4MCa4ZhYbIiJQNBQ2JQTvZIC0egQwIhBgB2f0BgNgQQX0csUcqcleCox2Hk%2BQZIF5eNFOR2NSC5MAzOJi4UL4JoU4e0BgdirVF4NiHi/guChCgQLmZC74boWyXsXVRk%2BipUTabcfUAQLC2Co4RQeqYKAi0yO0OsIo4C8YOjQ4TCGaK/JvOEGUmEMICaWWXSzOX8ycylXRfRKgNxExFeAAPw8vGC8pcQMSMX8qJUzhkpdjksRAUtUCUptlUtpHUpXH3VEHxDcsRNwsQtoDxmCmAGKOiDymMqIs2AIHSGzFUsnLo08NaObn4pYrkGSHAqEGQQOkY3qsmk1UqK9EwCEpErao6rByktATKNCOQUED3kkBkkOB7TAiXPrXPRaJ6s2SWoVjAiwAdi5n2y%2BMznCOblLQIC4Ezkzn/H8ETHml2UUMx0Op5lLFfHVLAnMK2qlHFOUoJA4nMMUl5HeogA0FmS4Gq3UkOH/A0F71fyBz%2BpCpYAIDj0XTQjQowryTsXAtmRDREDECaCxoMGIBWtN0EDclYG6laT0QWqWpn32sMThFoCoBdmAwngAFYrBRMmbTgXlJCBIhN851JzDZkfrua/kOCFASb6ACUpVZCabDZ9VGa6bRLIb0SaIwIPgXo5AjlYtaAIAKErd9qlbUIVa1ajkn8FDdbZD9b0LsAABFLkMWzAWZJ/DAxW6iA262220ee2w4OAnWz9RG1C9C64VG9edGwJEGXGnGxhJoAminIm0Wj2zdCm0hKmqWyEQ2eWzdEwFmlFdmzm36x8twfmrUXwoWrFDgmDd20mgjMFTHaW2g91OW%2Bm5OleC2jCokOCDW/TLWnW52jEgOtuk8bW2XU2nu5Wt2uO0mh22XfDBGsAokJGgOoO/OEOzGyO/GtfFA2Ou2hO9DRapOyG2u9OxmzO1mg%2BHOnkLmtkfOwuzZfkeMX64Wt/IFO2iWialeWu2W7y%2BWpui%2BFu1W9WiATWwe9eM25ul29Cv%2B42oe9A2nEe12m28e%2BgSe7vJ2zHFuseu22Zb2ho322e%2BelG4Kggfw5g8Otesncw4m%2BOxmxOnEb%2BoeA%2B%2BmjOrOtmjm8%2BvOuxdSG%2B3aIwAW4uuYoBQvMu5gjNBBzXfaz4lOuED%2BvRL%2B2Bvu9ugBzuoB/OEBn%2BsB1u%2BRk26By3WRhCeBjBuulg6enBwePBwOghoh7WDsKbekUh/hoechkR7eggXemh/e1O6qBho%2Bph0%2Blhw4C%2Bnm0fAuwQWZThl4Iwe%2BuxhnQRlRUJ%2BMOubuBEWgYouGSG8Rt%2B9xqRjxhW1BtR64fujuuLbunJ3u9RgezR4zFBvWtR9Bj2pBoPUHN7Z5ERlRvSRdfo20ZQfMNOSJmO20VQTHchx4ZosxTplBCAch1QHh20R4aNVQCAVQdYQ4R4CAGZytJjSWkx/2/BtGycjGtuIRPAJBB9de2u1QJxlx5amu9xvGKhneym3vOh9xw%2B7y4%2B7O3x/x/OzZR6o5yZWSdSUZ4gRSEJl%2BW%2B8Jku2Q0bR%2BnmPUHyMCb5trf5rpn%2BSZpZ1JjZi%2Bd%2BuxTdGR4p5WiBhRwphonR/F8pybbRq5uEIixm6dNlHWl0DA3FuB%2BZ2ZZIC%2BhZxly2m2x4Fli%2BmZ4x5yUxxemGEOptA5o56OzOU585%2B5ilw2G57y6hy5r4x52mzxl57xzO95thrRb64Jh6hgQ517PiQFjiAF//deyF6G8u0tJgWF%2BFgrM14dFFmZtF1%2BjFjJrFhu7JqpkpvJ%2BRwBopn15Wv1spqBipmB2Vu2TdGlspOl/DDl3R5lu2NlnRt27l5NxSPlj4v25GsxnZ8YPZsO1aVBCVqV255xmV6m656Vve4ZrOehhm9Vk%2BzV3O7mz5kF4toUX5k1yCJFgSYFyGUF4ACJiFsPd/CpW16SBQTt2DR1o5Z111yNzJnFoNg2/FgNolhNklsNsl/PVNm2lFx22nFNhNtNupmGel49zN/dpNiALB4Bk91dzllZ2ZO9nW3lmezZ3NoVq6EV4hmd0t9xs58ti52hsxWu%2BV8mu52t5V8Dp5tVvRV55h1ty%2B9h3VwhnmadjMdAbtxFlBc15A6JoJG12FrD1BPDk1%2Bd2ZF1vgqtyRz1z%2BxunRkNgpruzdp90pyB7vYe09g989wx6fVYR9826prl/jsqfDD9/lq5QV8xgJ4QO4hwQDuEYDhV6D1xyNyD%2Ba9TpV0IuD1VxtxDjVs%2Bvx7V3m01rpgd6yIdkdxlF2AFpNrN7ldFlVmWhj6Rpjrdo2gltj4B4l7z0lypkTkpt2w92XDAl0dljjs9gxsqCLpz0BkLvj2L9AYTxL0esTlLqT7N3BrZvN4O3Z0Oy6BoeOJTk5oDmtjTujuVyr3T/ghtxh5tkzj5tDoJjDud%2BlOzormGCABzhdqutJ91%2Bj9ebFzzjjljnzpRmGFpsxFuibwLiN6L5Lp/fDSLm99NlbwTl0BL8awedpw4AF04Sc0riCi%2BCZgZomoZ/a%2B68h5AHDtABOOtzOQ747%2B48ZomlFwZ2Z%2BZxZ5Z1Zv5O79abEb7VmqwLlauqKl79phtdRKHxT77QpU%2Bb70%2BNb6j45YAbbpJC0e74Hk1MH5z2esxAKuHk7ylLFQHh7wQS4LmgnyVN1q5chkn%2B46yCn4HzdANZV2T/NjDzG%2BHo79piVpnhwFnnHhOdn2fUdyVirkDytyXiD2rsDwvBrrxprrVtt1roXjD2JsFvhyXy1nFTXxz1JwLenhnTFkbr1xXxnNBm2pIA7rp/n%2BHkXoHhOY39ytR9dxRwNyX3%2BgLndoL63G3pNo9y3Nbl/IPjb8L2nTH8P0T4PuxNLwvCP8T1L69lw4VOPwG%2B3lBR3k753yn%2BG6TskLngrgt7rhTk73P%2B4wXh3174XlAUXomxmjniFnpxnMttTitmDuX6tmX7v/g/TtOhD5m1XlD%2BTgdNr2ZTXkd4/KGnFcuw3/rl/QbxXc3ugy3t3oPu3zX/P4Hzf3J/Jyb73wPg/jR/3xbn3uPsL7vVbqLy/pLl9lL4AxFGPsRqtd3h/rPnfhvl3wQT9gVvLj%2ByFDL0bM%2BIKxmZjcAIhkAPkdAJkSlxMgFA2AVPMpyH6GcR%2BbzMflfT1ba9h24La3ER1yQ4D4mZgRJsk3lyLtYOWcX%2BphBej0hMinRXnDoTb7aFYQxAG5swJ5xPQxGnA%2B6Gv3rqMdvWevLokiFP4D0N2fnZfsILm6H8Fu5LbgYzgei11Qcm6EPvnhfxPR/imVYcDD1tAnkYYZ5R5AswtY8C7oLdNCi9Ddq6FBcxxRvBHAwKWEu8QeZ6FsUIKMD0iQsb4C6Elx4hbq8glgbbntDqC%2Bc7/DQacBcEsC9B7cTeP/xk6AC5OQMH5JkCQGpxIB3QGAXAMYAGwCAyQkgCgKyaNcMBrDdXjq0n6YcuGuA3Xl1yZov0nu1A2gfQIWJMDJeD0WSOwKEFuDJBbgvgaDA36dCHovvf%2BuIOUZu8WBAwrjkHh45%2BDhB79FOCoKnqCcghTAvRFoIlA6DS86hTeOSiMHIFRhajcwZYJrzWCvc1%2BfDJYQfyOCL2zg2fu4PSKeDvB80GbjvmEEBCRQiwiHismCFhDGh6w08tEKL7EgS%2BS9QrgkIxSZAjuUscAW1gyEMAYMeQ55kZ1H5FDUOJQmFACys4gAbOeAgRlC2tYgtyhxA0gagBSa0c9OVAj3jQLoEMCNBzQ9jGwOMEKC%2BcUwroR6wt4CCreDOfoaINY5Tcrojw5QlIK5E8iyokw9oQ9BhQd89E97czjSKWEWhAoqwmFJENDibDDBErDkfdDMGq1LBleYQafnPz6F9EqxYwil3oIOChG6AYPOEKeFuCHodw1aA8JGHPC98bw4jB0P7hWi3BSojeKLhiHF84h3PPZokIEDgi2YygWSNCNhHlcDOBQ5DkiPH6ztLOZQsJhUPpEEDRkRA56ISOJFv9TeqjEph8ApENDtiwQmUawLaHXDGRoo3gSyPX5sjHRbg6Qf6y97scqxpgwUbIL3Z9DqxKnOYevDZYujNB8ow4GsK9EGDCk3wNUSYJEElN9hNtCsULCsF15jilBNmNQS5imEEiggcsIwSsLd4rhOom0fdDtE%2BC%2BR6ou6C8IHG8VQhHoh6KOL%2BE5dGk/xBEkcDXIgkYSgEKgPhAqgwQwI9Ia4IbWwC6QyQ5DPUGEHe6FEF4yAWZM2CaAiFT4GxLOuzQlbWR6KoFEAEyANYEAAA4vpiKqYBwK8TPeKfCgmelgAPWTHBwTVoRAwIbdYICrX9pCAwOQWWEvgCoAcB1gVkDgEzV4B%2BBuAvAVAJwAuKg83ImwbYGpFwg8BSAvwDgFoDpykBZQIAJmoDU4mSAeJmgfiZwF4AAxAa0k2SaQDgCwAYAiAFAI6TuBkAKAwSaMGZJADAAuAuEPgHQCrAAwAG6k0gNEDCC4xOAkk9ycwDYH0hog2gS6F5N4BRhqKBAekAwHyquSsAEIYAG4CKIAw%2BJpALAHqCMDiAZJvAfAEaB6DVJEpWgIIIlQtiCoMp5ALcZxPymnQgwbAjwFgFcm7QWAwU0gFqj0xKAjuqUoqqBQynrAqABgReB2A7Iih6Q08RqfwEEC412AUgGQIIAMpqBXJugJIAYCMAoA8e%2BgPaADFgAYwbJuE%2BgPRSakjA7iDAWUHJMaj4hEp1PLiMBlMCXFLAHQPxrQAnLjAwIoEhgPxPEpcV1pJcaoLUEyAuAcQUwdoEEBxCDBSg5QPQKkHSD4h/pYMgoPiGBnDAKgX0y6L0DmBQyxgNQJGfUDmBwylgCMiaJME8BtA9AeM%2BYCUHhkSB1gCgUSTsHJn6BOA3E0gLxPykCSOAhwVQAAA494lwGavNUjCHA7JLof8MONwCEBFQAENYLwF0nsT1gCkpSbTI4CqTSADU2WYzI0kcAtJIAHSepPWAGTjJFVaOIQwslRg/I9AYgBEBtScB2ZnM7mcAF5lSAXQkkoUCLIkp6BRpwgRhBNOkCuyZp6gEqfNNIAiggwyQYKRxLplqSSpzMyqLOEhCmVWZHMrmcCBtnNx%2BZgsiAB4CskmyV8/4cWVJK1nrAEQyoEYJ9JUm8AlZgNFWaQGZnqzNZXU6WYpOUmcB/wYcpmZpJzk1y5ZZgJuarIlm5y9pCZPwJICAA%3D).

include/up-cpp/utils/Expected.h
```diff
@@ -27,82 +28,137 @@ static_assert(!__has_cpp_attribute(__cpp_lib_expected),
...
+
+	constexpr explicit operator bool() const noexcept {
+		return std::holds_alternative<T>(storage);
+	}
+
+	// template <class X>
+	// constexpr T value_or(X&& v) const& noexcept {
+	// 	return has_value() ? std::get<T>(storage) :
+	// static_cast<E>(std::forward<X>(v));
+	// }
```
gregmedd, line 99:
```suggestion
```

include/up-cpp/utils/Expected.h
```diff
@@ -27,82 +28,137 @@ static_assert(!__has_cpp_attribute(__cpp_lib_expected),
...
+			    "Attempt to dereference expected pointer when unexpected.");
+		return &std::get<T>(storage);
+	}
 
 private:
+	std::variant<T, Unexpected<E>> storage;
+
 	static_assert(!std::is_void_v<T>,
-	              "We don't allow T==void (unlike std::expected)");
+	              "We don't allow E==void (unlike std::expected)");
```
gregmedd, line 152:
```suggestion
	              "We don't allow T==void (unlike std::expected)");
```

include/up-cpp/utils/Expected.h
```diff
@@ -27,82 +28,137 @@ static_assert(!__has_cpp_attribute(__cpp_lib_expected),
...
 
 	static_assert(
 	    std::is_destructible_v<T> && !std::is_array_v<T> &&
 	        !std::is_reference_v<T>,
-	    "Expected requires T to meet the C++ 'destructable' requirement");
+	    "Expected requires E to meet the C++ 'destructable' requirement");
```
gregmedd, line 157:
```suggestion
	    "Expected requires T to meet the C++ 'destructable' requirement");
```

test/coverage/utils/ExpectedTest.cpp
gregmedd: Could that not be proved by passing parameters to the regular constructors of `Expected`? It looks to me that the lambda is unnecessary indirection here.

What is the benefit of
```
int sample = get_rand();
auto lfn = [&]() -> Expected<int, int> { return Unexpected(sample); };
auto expected = lfn();
```
over
```
int sample = get_rand();
Expected<int, int> expected(Unexpected(sample));
```

gregmedd (commented):
Still some active discussions and minor changes.

test/coverage/utils/ExpectedTest.cpp
debruce: It seems to me mocking the function return behavior is closer to the real world behavior, and the syntax overhead is minimal. It may be the case that the lambda function is effectively inlined by the compiler. Who knows? I'm not sure that it matters. These unit tests are miles away from actually meaningful tests where integration is actually tested. I don't care either way. If you want them to not use lambda functions, I'll rip all the lambda functions out.

Is that the ask?

debruce: I'm assuming that you want the lambda functions ripped out. I will start to work on that. However, in tests like this, there is no meaningful 'optimization'. Unit tests don't prove much, and 'optimizing' seems like raking leaves.

debruce: In the three years I spent writing the testing framework for the ADIM3 interface to the A100, I never wrote a gtest. I used ROS/python Lunar Loggerhead interface, pybind11 for C++ encapsulation, pcapplusplus for packet assembly and spoofing, threads and condition variables for packet timing, OpenSSL for mTLS exchange, rpyc for dspace integration, and I was hoping to work on NeoVi Fire2 Linux libraries for live CAN capture.

My test specs were written in YAML, and were imported to python with ruamel.YAML. Pybind11 gave my C++ objects for packet description, timer chain objects, etc. python presence. I used nlohman::json for CBOR serialization and deserialization of variable length records, and fixed sized record index file for the ability to reel-rock the ethernet+ROS topic combined file stream.

I had written python code to parse the ARXML CAN message database, and use the enum values in the YAML driven testing specs.

I never considered gtests to be useful.

All that stuff I wrote is the main line stuff that Cruise is still using to test firmware drops. GM is also using it to test both A100 and A110 for ethernet behavior.

debruce: The thing I hated most about the Cruise situation was that Lunar Loggerhead was six years old, and didn't have a working message reflection interface. This is something the ROS folks fixed. I was hoping that such a thing would be fixed for this project, but that does not seem to be the case. I have code that proves it for all of the current uProtocol fields for everything but the details field. Fortunately, there seem to be no other repeated or any use cases, but even though it works and fills an obvious need, its dead code, never to be used.

gregmedd: > I'm assuming that you want the lambda functions ripped out. I will start to work on that. However, in tests like this, there is no meaningful 'optimization'. Unit tests don't prove much, and 'optimizing' seems like raking leaves.

@debruce - I want to be able to read the test and know what is being tested vs what is just extra stuff. It seems to me the lambdas are not adding anything here in terms of the testing. Instead they only add cognitive load for the reader. They might be useful if they were doing some sequence of tasks that needed to be reused throughout a test, but all they are doing is adding extra steps to constructing an object - that construction is the real purpose of the test.

include/up-cpp/utils/Expected.h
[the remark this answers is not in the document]
debruce: I thought the behavior we would want is to see that uint64_t to int is a reasonable type conversion (as long as it doesn't lose bits, which it might depending on int size for the compiler), and that was the purpose of the static_cast you suggested.
Making the type be different wasn't a bug. It was meant to test the tolerance for automatic type promotion.
I don't know if you consider flexible type promotion a virtue or not. That is a question to ask.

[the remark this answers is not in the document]
debruce: Is uint64_t to int64_t a reasonable automatic type promotion, or is if its from a const rvalue to an lvalue?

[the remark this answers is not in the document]
debruce: Regardless of implicit type promotion, the problem I saw was that there was an error coming from UTransport.cpp's use of Expected when the constructors were combined with a templated constructor. If you think that this is still possible, I will restore that situation and send you the error message I was getting.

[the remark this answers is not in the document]
debruce: This is the error message that comes about from trying to combine all the constructors into one templated constructor for the non-Unexpected case:

/home/gzy3g3/projects/up-cpp/src/transport/UTransport.cpp: In member function ‘uprotocol::utils::Expected<uprotocol::utils::callbacks::CalleeHandle<void, const uprotocol::v1::UMessage&>, uprotocol::v1::UStatus> uprotocol::transport::UTransport::registerListener(const uprotocol::v1::UUri&, uprotocol::transport::UTransport::ListenCallback&&, std::optional<uprotocol::v1::UUri>&&)’:
/home/gzy3g3/projects/up-cpp/src/transport/UTransport.cpp:69:33: error: could not convert ‘std::move<uprotocol::utils::callbacks::CalleeHandle<void, const uprotocol::v1::UMessage&>&>((* & handle))’ from ‘std::remove_reference<uprotocol::utils::callbacks::CalleeHandle<void, const uprotocol::v1::UMessage&>&>::type’ {aka ‘uprotocol::utils::callbacks::CalleeHandle<void, const uprotocol::v1::UMessage&>’} to ‘uprotocol::utils::Expected<uprotocol::utils::callbacks::CalleeHandle<void, const uprotocol::v1::UMessage&>, uprotocol::v1::UStatus>’
   69 |                 return std::move(handle);
      |                        ~~~~~~~~~^~~~~~~~
      |                                 |
      |                                 std::remove_reference<uprotocol::utils::callbacks::CalleeHandle<void, const uprotocol::v1::UMessage&>&>::type {aka uprotocol::utils::callbacks::CalleeHandle<void, const uprotocol::v1::UMessage&>}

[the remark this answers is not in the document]
debruce: The templated constructor that I'm trying to merge the prior indivdual constructors into is this:

	template <class X = T>
	constexpr explicit Expected(X&& rhs) {
		std::get<T>(storage_) = static_cast<T>(std::forward<X>(rhs));
	}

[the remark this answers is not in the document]
debruce: If you can recommend a fix for this, I'm happy to try it.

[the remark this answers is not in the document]
debruce: This pattern of multiple constructors and trying to debug stuff on the constructor signature could probably be solved in a more general, and much more readable way with one templated constructor and 'if constexpr ()' inside the constructor. Is that acceptable as a path forward?

include/up-cpp/utils/Expected.h
gregmedd: The suggested change is to remove the commented-out code. 5 deleted lines.

include/up-cpp/utils/Expected.h
```diff
@@ -27,82 +28,137 @@ static_assert(!__has_cpp_attribute(__cpp_lib_expected),
...
-	template <typename F>
-	constexpr auto or_else(F&& f) &;
-	template <typename F>
-	constexpr auto or_else(F&& f) const&;
-	template <typename F>
-	constexpr auto or_else(F&& f) &&;
-	template <typename F>
-	constexpr auto or_else(F&& f) const&&;
+class Expected {
+public:
```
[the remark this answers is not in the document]
gregmedd: @debruce - You may disagree with this, but for consistency with the current codebase _please_ make this change. This is not the first time I have made this _exact_ request in this PR.

Once a coding standard has been reviewed and adopted, this may change.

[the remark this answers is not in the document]
gregmedd: There are things I care more about in this review that have not been addressed. This can be evaluated when the coding standard is adopted.

gregmedd (approved):
LGTM - thanks for getting this through review!

gregmedd:
@debruce - Do you want to squash and rebase this commit? Otherwise I'll just do the automatic squash and merge in GitHub

PR #128, opened 2024-05-29
Add UUID Validator and its unit tests

BishrutSubedi:
This PR intends to add UUID validator, and it's unit tests in eclipse up-cpp

### DIFF
--- a/src/datamodel/validator/Uuid.cpp
+++ b/src/datamodel/validator/Uuid.cpp
@@ -10,3 +10,145 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "up-cpp/datamodel/validator/Uuid.h"
+
+using namespace std::chrono_literals;
+
+namespace {
+
+constexpr uint64_t TIMESTAMP_MASK = 0xFFFFFFFFFFFF;
+constexpr uint64_t TIMESTAMP_SHIFT = 16;
+constexpr uint64_t VERSION_MASK = 0xF;
+constexpr uint64_t VERSION_SHIFT = 12;
+constexpr uint64_t VARIANT_MASK = 0x3;
+constexpr uint64_t VARIANT_SHIFT = 62;
+constexpr uint64_t COUNTER_MASK = 0xFFF;
+
+using milliseconds = std::chrono::milliseconds;
+
+std::chrono::system_clock::time_point getUuidTimestamp(
+    const uprotocol::v1::UUID& uuid) {
+	uint64_t msb = uuid.msb();
+	uint64_t timestamp = (msb >> TIMESTAMP_SHIFT) & TIMESTAMP_MASK;
+
+	return std::chrono::system_clock::time_point(milliseconds(timestamp));
+}
+
+uint8_t internalGetVersion(const uprotocol::v1::UUID& uuid) {
+	return (uuid.msb() >> VERSION_SHIFT) & VERSION_MASK;
+}
+
+uint8_t internalGetVariant(const uprotocol::v1::UUID& uuid) {
+	return (uuid.lsb() >> VARIANT_SHIFT) & VARIANT_MASK;
+}
+
+}  // namespace
+
+namespace uprotocol::datamodel::validator::uuid {
+
+std::string_view message(Reason reason) {
+	switch (reason) {
+		case Reason::WRONG_VERSION:
+			return "Invalid UUID version";
+		case Reason::UNSUPPORTED_VARIANT:
+			return "Unsupported UUID variant";
+		case Reason::FROM_THE_FUTURE:
+			return "UUID timestamp is from the future";
+		case Reason::EXPIRED:
+			return "UUID has expired";
+		default:
+			return "Unknown reason";
+	}
+}
+
+ValidationResult isUuid(const uprotocol::v1::UUID uuid) {
+	uint64_t msb = uuid.msb();
+	uint64_t lsb = uuid.lsb();
+
+	uint8_t version = internalGetVersion(uuid);
+	if (version != 8) {
+		return {false, Reason::WRONG_VERSION};
+	}
+
+	uint8_t variant = internalGetVariant(uuid);
+	if (variant != 2) {  // Variant should be 10 in binary
+		return {false, Reason::UNSUPPORTED_VARIANT};
+	}
+
+	auto timestamp = getUuidTimestamp(uuid);
+	auto current_time = std::chrono::system_clock::now();
+
+	if (timestamp > current_time) {
+		return {false, Reason::FROM_THE_FUTURE};
+	}
+
+	return {true, std::nullopt};
+}
+
+ValidationResult isExpired(const uprotocol::v1::UUID uuid,
+                           std::chrono::milliseconds ttl) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		return {false, reason};
+	}
+
+	auto timestamp = getUuidTimestamp(uuid);
+	auto current_time = std::chrono::system_clock::now();
+
+	if ((current_time - timestamp) > ttl) {
+		return {true, Reason::EXPIRED};
+	}
+
+	return {false, std::nullopt};
+}
+
+uint8_t getVersion(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+	return internalGetVersion(uuid);
+}
+
+uint8_t getVariant(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+	return internalGetVariant(uuid);
+}
+
+std::chrono::system_clock::time_point getTime(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+	return getUuidTimestamp(uuid);
+}
+
+std::chrono::milliseconds getElapsedTime(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+
+	auto current_time = std::chrono::system_clock::now();
+	auto uuid_time = getTime(uuid);
+
+	return std::chrono::duration_cast<milliseconds>(current_time - uuid_time);
+}
+
+std::chrono::milliseconds getRemainingTime(const uprotocol::v1::UUID uuid,
+                                           std::chrono::milliseconds ttl) {
+	auto elapsed_time = getElapsedTime(uuid);
+	return std::max(ttl - elapsed_time, 0ms);
+}
+
+uint16_t getCounter(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+	return (uuid.msb() & COUNTER_MASK);
+}
+
+}  // namespace uprotocol::datamodel::validator::uuid
--- a/test/coverage/datamodel/UuidValidatorTest.cpp
+++ b/test/coverage/datamodel/UuidValidatorTest.cpp
@@ -11,10 +11,16 @@
 
 #include <gtest/gtest.h>
 #include <up-cpp/datamodel/validator/Uuid.h>
+#include <uprotocol/v1/uuid.pb.h>
+
+#include <chrono>
+
+using namespace uprotocol::datamodel;
+using namespace std::chrono_literals;
 
 namespace {
 
-class TestFixture : public testing::Test {
+class TestUuidValidator : public testing::Test {
 protected:
 	// Run once per TEST_F.
 	// Used to set up clean environments per test.
@@ -23,8 +29,8 @@ protected:
 
 	// Run once per execution of the test application.
 	// Used for setup of all tests. Has access to this instance.
-	TestFixture() = default;
-	~TestFixture() = default;
+	TestUuidValidator() = default;
+	~TestUuidValidator() = default;
 
 	// Run once per execution of the test application.
 	// Used only for global setup outside of tests.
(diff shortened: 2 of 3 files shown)


src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,139 @@
...
+		return {false, Reason::EXPIRED};
+	}
+
+	return {true, std::nullopt};
+}
+
+uint8_t getVersion(const uprotocol::v1::UUID uuid) {
+	auto [valid, maybe_reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid("Invalid UUID version");
```
gregmedd, line 89: This should probably throw the reason. The UUID might not be invalid because of the version.

```suggestion
		throw InvalidUuid(message(maybe_reason.value_or("Unknown reason")));
```

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,139 @@
...
+	if (!valid) {
+		throw InvalidUuid("Invalid UUID counter");
+	}
+	return (uuid.msb() & 0xFFF);  // Extract 12-bit counter
+}
+
+InvalidUuid::InvalidUuid(InvalidUuid&&) noexcept = default;
+InvalidUuid& InvalidUuid::operator=(InvalidUuid&&) noexcept = default;
+InvalidUuid::InvalidUuid(const InvalidUuid&) = default;
+InvalidUuid& InvalidUuid::operator=(const InvalidUuid&) = default;
```
gregmedd, line 146: It's fine to put `= default` in the .h file

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,139 @@
...
+	if (variant != 2) {  // Variant should be 10 in binary
+		return {false, Reason::UNSUPPORTED_VARIANT};
+	}
+
+	uint64_t timestamp =
+	    (msb >> 16) & 0xFFFFFFFFFFFF;  // Extract 48-bit timestamp
+	uint64_t current_time =
+	    std::chrono::duration_cast<std::chrono::milliseconds>(
+	        std::chrono::system_clock::now().time_since_epoch())
+	        .count();
```
gregmedd, line 58: I would generally recommend against using `.count()` in this context. It would be safer to convert the timestamp from the UUID to a time point:
```suggestion
	std::chrono::milliseconds uuid_time(
	    (msb >> 16) & 0xFFFFFFFFFFFF);  // Extract 48-bit timestamp
	std::chrono::system_clock::time_point timestamp(uuid_time);
	auto current_time = std::chrono::system_clock::now();
```

gregmedd: It might also make sense to put a `getUuidTimestamp()` method in the anonymous namespace of this file so the code doesn't need to be repeated for each function

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,139 @@
...
+	if (!valid) {
+		throw InvalidUuid("Invalid UUID timestamp");
+	}
+
+	uint64_t current_time =
+	    std::chrono::duration_cast<std::chrono::milliseconds>(
+	        std::chrono::system_clock::now().time_since_epoch())
+	        .count();
+	uint64_t uuid_time = (uuid.msb() >> 16) & 0xFFFFFFFFFFFF;
+	return std::chrono::milliseconds(current_time - uuid_time);
```
gregmedd, line 123: This should use `getTime()` to get the time from the UUID, then do all the comparisons and arithmetic on the chrono objects.
```suggestion
	auto current_time = std::chrono::system_clock::now();
	auto uuid_time = getTime(uuid);
	return std::chrono::duration_cast<std::chrono::milliseconds>(current_time - uuid_time);
```

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,139 @@
...
+}
+
+std::chrono::milliseconds getRemainingTime(const uprotocol::v1::UUID uuid,
+                                           std::chrono::milliseconds ttl) {
+	auto elapsed_time = getElapsedTime(uuid);
+	if (elapsed_time > ttl) {
+		return std::chrono::milliseconds(0);
+	}
+	return ttl - elapsed_time;
```
gregmedd, line 132: Minor: This could be condensed
```suggestion
	auto elapsed_time = getElapsedTime(uuid);
	return std::min(ttl - elapsed_time, 0ms);
```

test/coverage/datamodel/UuidValidatorTest.cpp
```diff
@@ -23,16 +26,216 @@ class TestFixture : public testing::Test {
...
+}
+
+// Test UUID from the future
+TEST_F(TestUuidValidator, FromTheFuture) {
+	// Creating UUID with a timestamp in the future
+	uint64_t future_timestamp =
+	    std::chrono::duration_cast<std::chrono::milliseconds>(
+	        std::chrono::system_clock::now().time_since_epoch())
+	        .count() +
+	    100000;  // 100 seconds in the future
```
gregmedd, line 95: I would recommend doing the time arithmetic on chrono objects before converting via `.count()`
```suggestion
	auto future_time = std::chrono::system_clock::now() + 100s;
	uint64_t future_timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(
	    future_time.time_since_epoch()).count();
```

test/coverage/datamodel/UuidValidatorTest.cpp
```diff
@@ -23,16 +26,216 @@ class TestFixture : public testing::Test {
...
+	        std::chrono::system_clock::now().time_since_epoch())
+	        .count() -
+	    100000;  // 100 seconds in the past
+	uint64_t msb = (8ULL << 12) | (past_timestamp << 16) |
+	               (0x123ULL);  // Version 8, past timestamp
+	uint64_t lsb = (2ULL << 62) | (0xFFFFFFFFFFFFULL);  // Variant 10
+
+	uprotocol::v1::UUID uuid = createFakeUuid(msb, lsb);
+
+	auto [valid, reason] = uprotocol::datamodel::validator::uuid::isExpired(
```
gregmedd, line 122: Minor: It would be reasonable to add a `using namespace uprotocol::datamodel;` statement in this .cpp file to shorten some of these identifiers.

test/coverage/datamodel/UuidValidatorTest.cpp
```diff
@@ -23,16 +26,216 @@ class TestFixture : public testing::Test {
...
+	    uuid, std::chrono::milliseconds(60000));  // 60 seconds TTL
```
gregmedd, line 123: Use chrono literals: `60s` is easier to read than `std::chrono::milliseconds(60000)`

test/coverage/datamodel/UuidValidatorTest.cpp
```diff
@@ -23,16 +26,216 @@ class TestFixture : public testing::Test {
...
+	    (8ULL << 12) | (timestamp << 16) | (0x123ULL);  // Version 8, timestamp
+	uint64_t lsb = (2ULL << 62) | (0xFFFFFFFFFFFFULL);  // Variant 10
+
+	uprotocol::v1::UUID uuid = createFakeUuid(msb, lsb);
+
+	auto uuid_time = uprotocol::datamodel::validator::uuid::getTime(uuid);
+	auto uuid_timestamp = std::chrono::duration_cast<std::chrono::milliseconds>(
+	                          uuid_time.time_since_epoch())
+	                          .count();
+	EXPECT_EQ(uuid_timestamp, timestamp);
```
gregmedd, line 201: See earlier comments about doing time arithmetic and comparisons using chrono objects

test/coverage/datamodel/UuidValidatorTest.cpp
```diff
@@ -23,16 +26,216 @@ class TestFixture : public testing::Test {
...
+	        std::chrono::system_clock::now().time_since_epoch())
+	        .count() -
+	    30000;  // 30 seconds in the past
+	uint64_t msb =
+	    (8ULL << 12) | (timestamp << 16) | (0x123ULL);  // Version 8, timestamp
+	uint64_t lsb = (2ULL << 62) | (0xFFFFFFFFFFFFULL);  // Variant 10
+
+	uprotocol::v1::UUID uuid = createFakeUuid(msb, lsb);
+
+	EXPECT_GE(
```
gregmedd, line 217: How much greater than 30s is allowed?

test/coverage/datamodel/UuidValidatorTest.cpp
```diff
@@ -23,16 +26,216 @@ class TestFixture : public testing::Test {
...
+	        std::chrono::system_clock::now().time_since_epoch())
+	        .count() -
+	    30000;  // 30 seconds in the past
+	uint64_t msb =
+	    (8ULL << 12) | (timestamp << 16) | (0x123ULL);  // Version 8, timestamp
+	uint64_t lsb = (2ULL << 62) | (0xFFFFFFFFFFFFULL);  // Variant 10
+
+	uprotocol::v1::UUID uuid = createFakeUuid(msb, lsb);
+
+	EXPECT_LE(uprotocol::datamodel::validator::uuid::getRemainingTime(
```
gregmedd, line 235: How much less than 30s is allowed?

test/coverage/datamodel/UuidValidatorTest.cpp
gregmedd, line 1: Add a few `EXPECT_THROW` test cases in for situations where an exception is expected to be thrown.

gregmedd (changes requested):

test/coverage/datamodel/UuidValidatorTest.cpp
BishrutSubedi: sounds good, I'll update this with 10 second tolerance

test/coverage/datamodel/UuidValidatorTest.cpp
BishrutSubedi: Thanks, I'll update this with 10 second tolerance

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,134 @@
...
+}
+
+std::chrono::system_clock::time_point getTime(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+	uint64_t timestamp = (uuid.msb() >> 16) & 0xFFFFFFFFFFFF;
+	return std::chrono::system_clock::time_point(
+	    std::chrono::milliseconds(timestamp));
```
gregmedd, line 112:
```suggestion
	return getUuidTimestamp(uuid);
```

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,134 @@
...
+
+std::chrono::milliseconds getElapsedTime(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+
+	std::chrono::system_clock::time_point current_time =
+	    std::chrono::system_clock::now();
+	std::chrono::system_clock::time_point uuid_time = getTime(uuid);
```
gregmedd, line 123:
```suggestion
	auto current_time = std::chrono::system_clock::now();
	auto uuid_time = getTime(uuid);
```

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,134 @@
...
+
+	return std::chrono::duration_cast<std::chrono::milliseconds>(current_time -
+	                                                             uuid_time);
```
gregmedd, line 126: Minor: These are pretty long. You might consider adding
```
template <typename T>
using duration_cast = std::chrono::duration_cast<T>;
using milliseconds = std::chrono::milliseconds;
```
in the anonymous namespace.

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,134 @@
...
 #include "up-cpp/datamodel/validator/Uuid.h"
+
+using namespace std::chrono_literals;
+
+namespace {
+std::chrono::system_clock::time_point getUuidTimestamp(
+    const uprotocol::v1::UUID& uuid) {
+	uint64_t msb = uuid.msb();
+	uint64_t timestamp =
+	    (msb >> 16) & 0xFFFFFFFFFFFF;  // Extract 48-bit timestamp
```
billpittman, line 21: I'd recommend using constants (or constexpr's) for a lot of these bitmasks and shift values.  Alternatively create a macro that you can feed the relavant pieces into.

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,134 @@
...
+		default:
+			return "Unknown reason";
+	}
+}
+
+ValidationResult isUuid(const uprotocol::v1::UUID uuid) {
+	uint64_t msb = uuid.msb();
+	uint64_t lsb = uuid.lsb();
+
+	uint8_t version = (msb >> 12) & 0xF;
```
billpittman, line 48: You've got this same code several places.  It might be easier to read if you created an _internal_getVersion that only does the shifting/bitmasking, but no additional checks, then both isUuid and getVersion could both call the new function.

src/datamodel/validator/Uuid.cpp
```diff
@@ -10,3 +10,134 @@
...
+	}
+	return (uuid.msb() >> 12) & 0xF;
+}
+
+uint8_t getVariant(const uprotocol::v1::UUID uuid) {
+	auto [valid, reason] = isUuid(uuid);
+	if (!valid) {
+		throw InvalidUuid(message(reason.value()));
+	}
+	return (uuid.lsb() >> 62) & 0x3;
```
billpittman, line 102: Consider an _internal_getVariant method that only does the shifting/bitmasking.

billpittman (changes requested):
Overall looks good, when dealing with bitmasks and shift values, I tend to try to single source that code as much as possible in the hopes of reducing mistakes.  See inline comments.

src/datamodel/validator/Uuid.cpp
BishrutSubedi: Thank you, I'll add these suggestions, and following. It should improve readability

billpittman (approved):
Thanks for the changes, looks great!

PR #139, opened 2024-06-01
UUID serializer and its unit test

BishrutSubedi:
This PR adds UUID serializer for uprotocol uuid v8  #122

### DIFF
--- a/src/datamodel/serializer/Uuid.cpp
+++ b/src/datamodel/serializer/Uuid.cpp
@@ -10,3 +10,154 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "up-cpp/datamodel/serializer/Uuid.h"
+
+#include <arpa/inet.h>
+
+#include <iomanip>
+#include <sstream>
+#include <stdexcept>
+
+namespace {
+constexpr uint64_t TIMESTAMP_MASK = 0xFFFFFFFFFFFF;
+constexpr uint64_t TIMESTAMP_SHIFT = 16;
+constexpr uint64_t VERSION_MASK = 0xF;
+constexpr uint64_t VERSION_SHIFT = 12;
+constexpr uint64_t COUNTER_MASK = 0xFFF;
+constexpr uint64_t VARIANT_MASK = 0x3;
+constexpr uint64_t VARIANT_SHIFT = 62;
+constexpr uint64_t RANDOM_MASK = 0x3FFFFFFFFFFFFFFF;
+constexpr uint64_t RANDOM_SHIFT = 48;
+constexpr size_t UUID_BYTE_SIZE = 16;
+constexpr size_t UUID_PART_SIZE = 4;
+constexpr uint32_t HEX_BASE = 16;
+constexpr uint64_t MASK_32_BITS = 0xFFFFFFFF;
+constexpr uint64_t MASK_16_BITS = 0xFFFF;
+constexpr uint64_t MASK_14_BITS = 0x3FFF;
+constexpr size_t MSB_HIGH_ = 0;
+constexpr size_t MSB_LOW__ = 4;
+constexpr size_t LSB_HIGH_ = 8;
+constexpr size_t LSB_LOW__ = 12;
+}  // namespace
+
+namespace uprotocol::datamodel::serializer::uuid {
+
+std::string AsString::serialize(const uprotocol::v1::UUID uuid) {
+	// Extracting the parts of the UUIDv8
+	uint64_t unix_ts_ms = (uuid.msb() >> TIMESTAMP_SHIFT) & TIMESTAMP_MASK;
+	uint8_t ver = (uuid.msb() >> VERSION_SHIFT) & VERSION_MASK;
+	uint16_t counter = uuid.msb() & COUNTER_MASK;
+	uint8_t var = (uuid.lsb() >> VARIANT_SHIFT) & VARIANT_MASK;
+	uint64_t rand_b = uuid.lsb() & RANDOM_MASK;
+
+	// Formatting the UUIDv8 in the traditional format
+	std::stringstream ss;
+	ss << std::hex << std::setfill('0') << std::setw(8)
+	   << ((unix_ts_ms >> 16) & MASK_32_BITS)  // First 32 bits of timestamp
+	   << "-" << std::setw(4)
+	   << ((unix_ts_ms)&MASK_16_BITS)  // Next 16 bits of timestamp i.e. last 16
+	                                   // bit of ts
+	   << "-" << std::setw(4)
+	   << (((ver & VERSION_MASK) << 12) |
+	       (counter & COUNTER_MASK))  // Last 16 bits of timestamp and version
+	   << "-" << std::setw(4)
+	   << (((var & VARIANT_MASK) << 14) |
+	       ((rand_b >> RANDOM_SHIFT) & MASK_14_BITS))  // Variant and randb
+	   << "-" << std::setw(12) << (rand_b & TIMESTAMP_MASK);  // Random number
+
+	return std::move(ss).str();
+}
+
+uprotocol::v1::UUID AsString::deserialize(const std::string& str) {
+	// Check if the UUID string is in the correct format
+	// Format  : 12345678-1234-1234-1234-123456789012
+	// Index   : 01234567890123456789012345678901234
+	// Layout  : ***msb**-lsb*-vcnt-varr-***RAND*****
+	// msb - timestamp most significant bits (32 bits)
+	// lsb - timestamp least significant bits (16 bits)
+	// v - version (4 bits)
+	// cnt - counter (12 bits)
+	// var - variant (2 bits)
+	// RAND - random (62 bits)
+	// Please check UP-spec for UUID formatting:
+	// https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uuid.adoc
+
+	if (str.length() != 36 || str[8] != '-' || str[13] != '-' ||
+	    str[18] != '-' || str[23] != '-') {
+		throw std::invalid_argument("Invalid UUID string format");
+	}
+
+	uprotocol::v1::UUID uuid;
+	uint64_t unix_ts_ms = 0;
+	uint8_t ver = 0;
+	uint16_t counter = 0;
+	uint8_t var = 0;
+	uint64_t rand_b = 0;
+
+	try {
+		// Extract the parts from the UUID string
+		unix_ts_ms = std::stoull(str.substr(0, 8), nullptr, HEX_BASE) << 16;
+		unix_ts_ms |= std::stoull(str.substr(9, 4), nullptr, HEX_BASE);
+
+		uint16_t ver_counter = std::stoul(str.substr(14, 4), nullptr, HEX_BASE);
+		ver = (ver_counter >> 12) & VERSION_MASK;
+		counter = ver_counter & COUNTER_MASK;
+
+		uint16_t var_randb = std::stoul(str.substr(19, 4), nullptr, HEX_BASE);
+		var = (var_randb >> 14) & VARIANT_MASK;
+		rand_b = static_cast<uint64_t>(var_randb & MASK_14_BITS)
+		         << RANDOM_SHIFT;
+		rand_b |= std::stoull(str.substr(24), nullptr, HEX_BASE);
+	} catch (const std::exception& e) {
+		throw std::invalid_argument("Invalid UUID string format");
+	}
+
+	// Reconstruct the UUID
+	uuid.set_msb((unix_ts_ms << TIMESTAMP_SHIFT) | (ver << VERSION_SHIFT) |
+	             counter);
+	uuid.set_lsb((static_cast<uint64_t>(var) << VARIANT_SHIFT) | rand_b);
+
+	return uuid;
+}
+
+// Serialization function
+std::vector<uint8_t> AsBytes::serialize(const v1::UUID uuid) {
+	std::vector<uint8_t> bytes(UUID_BYTE_SIZE);
+
+	uint32_t msb_high = htonl(static_cast<uint32_t>(uuid.msb() >> 32));
+	uint32_t msb_low = htonl(static_cast<uint32_t>(uuid.msb() & MASK_32_BITS));
+	uint32_t lsb_high = htonl(static_cast<uint32_t>(uuid.lsb() >> 32));
+	uint32_t lsb_low = htonl(static_cast<uint32_t>(uuid.lsb() & MASK_32_BITS));
+
+	std::memcpy(&bytes[MSB_HIGH_], &msb_high, UUID_PART_SIZE);
+	std::memcpy(&bytes[MSB_LOW__], &msb_low, UUID_PART_SIZE);
+	std::memcpy(&bytes[LSB_HIGH_], &lsb_high, UUID_PART_SIZE);
+	std::memcpy(&bytes[LSB_LOW__], &lsb_low, UUID_PART_SIZE);
+
+	return bytes;
+}
+
+v1::UUID AsBytes::deserialize(const std::vector<uint8_t>& bytes) {
+	if (bytes.size() != UUID_BYTE_SIZE) {
+		throw std::invalid_argument("Invalid UUID byte array size");
+	}
+
+	uint32_t msb_high, msb_low, lsb_high, lsb_low;
+
+	std::memcpy(&msb_high, &bytes[MSB_HIGH_], UUID_PART_SIZE);
+	std::memcpy(&msb_low, &bytes[MSB_LOW__], UUID_PART_SIZE);
+	std::memcpy(&lsb_high, &bytes[LSB_HIGH_], UUID_PART_SIZE);
+	std::memcpy(&lsb_low, &bytes[LSB_LOW__], UUID_PART_SIZE);
+
+	uint64_t msb =
+	    (static_cast<uint64_t>(ntohl(msb_high)) << 32) | ntohl(msb_low);
+	uint64_t lsb =
+	    (static_cast<uint64_t>(ntohl(lsb_high)) << 32) | ntohl(lsb_low);
+
+	v1::UUID uuid;
+	uuid.set_msb(msb);
+	uuid.set_lsb(lsb);
+
+	return uuid;
+}
+
+}  // namespace uprotocol::datamodel::serializer::uuid
--- a/test/coverage/datamodel/UuidSerializerTest.cpp
+++ b/test/coverage/datamodel/UuidSerializerTest.cpp
@@ -11,10 +11,14 @@
 
 #include <gtest/gtest.h>
 #include <up-cpp/datamodel/serializer/Uuid.h>
+#include <uprotocol/v1/uuid.pb.h>
+
+#include <thread>
+#include <vector>
 
 namespace {
 
-class TestFixture : public testing::Test {
+class TestUuidSerializer : public testing::Test {
 protected:
 	// Run once per TEST_F.
 	// Used to set up clean environments per test.
@@ -23,8 +27,8 @@ protected:
 
 	// Run once per execution of the test application.
 	// Used for setup of all tests. Has access to this instance.
-	TestFixture() = default;
-	~TestFixture() = default;
+	TestUuidSerializer() = default;
+	~TestUuidSerializer() = default;
 
 	// Run once per execution of the test application.
 	// Used only for global setup outside of tests.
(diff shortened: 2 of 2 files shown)


src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,65 @@
...
+#include <sstream>
+#include <stdexcept>
+
+namespace uprotocol::datamodel::serializer::uuid {
+
+std::string AsString::serialize(const uprotocol::v1::UUID uuid) {
+	std::stringstream ss;
+	ss << std::hex << std::setw(16) << std::setfill('0') << uuid.msb() << "-"
+	   << std::setw(16) << std::setfill('0') << uuid.lsb();
+	return ss.str();
```
gregmedd, line 24: Changing this to `return std::move(ss).str();` tells the compiler that `ss` is no longer needed, allowing the returned string to be move constructed. This saves an allocate, copy, deallocate cycle (so long as the string is longer than the _small string optimization_ length of a given build environment (around 12-26 characters, depending on various factors).

```suggestion
	return std::move(ss).str();
```

Note that this is only needed here because the function returns the output of a method on `ss`. Normally the compiler can figure this out itself if directly returning a local variable (via _return value optimization_).

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,159 @@ class TestFixture : public testing::Test {
...
+	EXPECT_EQ(uuid.lsb(), 0xFEDCBA0987654321);
+}
+
+// Test invalid string deserialization
+// Two hyphen invalid string, 16-16 is valid uuid string
+TEST_F(TestUuidSerializer, DeserializeInvalidString) {
+	std::string invalid_uuid_str = "invalid-uuid-string";
+	EXPECT_THROW(uprotocol::datamodel::serializer::uuid::AsString::deserialize(
+	                 invalid_uuid_str),
+	             std::invalid_argument);
```
gregmedd, line 65: It might be worth adding a few more test cases here. For example, a zero-length string, a string that is the right length and has the `-` in the right place, but one character isn't valid hex (e.g. put a `z` in there)

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,159 @@ class TestFixture : public testing::Test {
...
+}
+
+// Test string deserialization
+TEST_F(TestUuidSerializer, DeserializeFromString) {
+	std::string uuid_str = "1234567890abcdef-fedcba0987654321";
+	uprotocol::v1::UUID uuid =
+	    uprotocol::datamodel::serializer::uuid::AsString::deserialize(uuid_str);
+
+	EXPECT_EQ(uuid.msb(), 0x1234567890ABCDEF);
+	EXPECT_EQ(uuid.lsb(), 0xFEDCBA0987654321);
```
gregmedd, line 56: It wouldn't hurt to test a couple more values, such as one with mixed case letters and one with some leading zeros.

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,159 @@ class TestFixture : public testing::Test {
...
 };
+// Test string serialization
+TEST_F(TestUuidSerializer, SerializeToString) {
+	uprotocol::v1::UUID uuid;
+	uuid.set_msb(0x1234567890ABCDEF);
+	uuid.set_lsb(0xFEDCBA0987654321);
+
+	std::string uuid_str =
+	    uprotocol::datamodel::serializer::uuid::AsString::serialize(uuid);
+	EXPECT_EQ(uuid_str, "1234567890abcdef-fedcba0987654321");
```
gregmedd, line 46: It wouldn't hurt to add a few extra test values, such as one that might produce leading zeros in each of the segments.

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,159 @@ class TestFixture : public testing::Test {
...
+	EXPECT_EQ(uuid_bytes, expected_bytes);
 
-// TODO replace
-TEST_F(TestFixture, SomeTestName) {}
+	deserialized_uuid =
+	    uprotocol::datamodel::serializer::uuid::AsBytes::deserialize(
+	        uuid_bytes);
+	EXPECT_EQ(deserialized_uuid.msb(), 0x0000000000000000);
+	EXPECT_EQ(deserialized_uuid.lsb(), 0x0000000000000000);
+}
```
gregmedd, line 183: Is this a duplicate of the minimum value test case?

gregmedd (changes requested):
For the most part, this looks good. I left a comment about the string format with a link to the relevant RFC, and suggested some additional test cases. Let me know if you have any questions.

test/coverage/datamodel/UuidSerializerTest.cpp
BishrutSubedi: yes, thanks for the catch

src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,65 @@
...
+std::vector<uint8_t> AsBytes::serialize(const v1::UUID uuid) {
+	std::vector<uint8_t> bytes(16);
+	uint64_t msb = uuid.msb();
+	uint64_t lsb = uuid.lsb();
+
+	// Copy msb and lsb into the byte vector in big-endian order
+	for (int i = 0; i < 8; ++i) {
+		bytes[i] = (msb >> (8 * (7 - i))) & 0xFF;
+		bytes[8 + i] = (lsb >> (8 * (7 - i))) & 0xFF;
+	}
```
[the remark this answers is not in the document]
BishrutSubedi: Thanks, I've updated this to use htonl and ntohl. It should work for either of endianness

stevenhartley (changes requested):

src/datamodel/serializer/Uuid.cpp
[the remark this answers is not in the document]
BishrutSubedi: Yes, sounds good. This serializer follows the 4-2-2-2-6 hextOctet format per RFC9562

The formatting part placing those bits in correct order. This could look bit different is our Uuid is custom defined, but those bits are placed correctly in serializer as we expect the uuid to look. I'm thinking this adds to readability as it is easier to see what fields are in UUID and the reason behind formatting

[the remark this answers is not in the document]
gregmedd: The variable names are version specific, but the actual conversion function looks to be version agnostic. It takes all the same bits and puts them in all the same places.

src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,122 @@
...
+	    (static_cast<uint64_t>(ntohl(msb_high)) << 32) | ntohl(msb_low);
+	uint64_t lsb =
+	    (static_cast<uint64_t>(ntohl(lsb_high)) << 32) | ntohl(lsb_low);
+
+	v1::UUID uuid;
+	uuid.set_msb(msb);
+	uuid.set_lsb(lsb);
+
+	return uuid;
+}
```
gregmedd, line 129: The more I think about it, the more I start to think we might not need this at all. Converting to/from a string makes a lot of sense for debugging or as a developer interface for loading data, but I'm having trouble coming up with an actual use case for the byte serialization.

I know there was a reason I put this in here back when I did the API, but that might have been a remnant of the way the old API worked.

We can keep it here for now and consider removing it later if there's no need for it.

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,222 @@ class TestFixture : public testing::Test {
...
+	// Assert the serialized UUID matches the expected value
+	EXPECT_EQ(serialized_uuid, "00001234-5678-90ab-fedc-ba0987654982");
+}
+
+// Test serialization with leading/trailing zeros and mixed case letters
+TEST_F(TestUuidSerializer,
+       SerializeWithLeadingZerosAndTrailingZerosAndMixedCaseLetters) {
+	uprotocol::v1::UUID uuid;
+	uuid.set_msb(0x00001234567890AB);
+	uuid.set_lsb(0xFedcba0987600000);
```
gregmedd, line 92: The mixed case letters should be in a string that is deserialized. Otherwise this test case is perfect for the leading zeros.

gregmedd (approved):
I'm ok with this. There is one small note on a test case.

I don't think we should hold up the PR while we decide if we need the byte serializer.

gregmedd (changes requested):
Ok, I actually have just one more change to ask for. Can you rebase this onto main (make sure it is the latest main from eclipse-uprotocol/up-cpp and not your fork) so that the CI can run?

While you are doing that, I would also appreciate it if you could use interactive rebase to squash your commits and clean up the commit messages. You don't have to squash down to one commit - whatever number clearly shows the units of work.

[This article provides some good guidance on commit messages](https://cbea.ms/git-commit/).

BishrutSubedi:
sounds good, I will squash to 1 commit with descriptive commit msg

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,235 @@ class TestFixture : public testing::Test {
...
+	    uprotocol::datamodel::serializer::uuid::AsString::deserialize(uuid_str);
+	// Assert the deserialized UUID matches the expected values
+	EXPECT_EQ(deserialized_uuid.msb(), 0x123456789ABCDEF0);
+	EXPECT_EQ(deserialized_uuid.lsb(), 0xFEDCBA9876543210);
+}
+
+// Test deserialization with leading/trailing zeros and mixed case letters
+TEST_F(TestUuidSerializer,
+       DeserializeWithLeadingZerosAndTrailingZerosAndMixedCaseLetters) {
+	std::string uuid_str = "00001234-5678-90ab-fedc-ba0987600000";
```
gregmedd, line 115: The mixed case letters should be in here. They don't matter in the integer literals
```suggestion
	std::string uuid_str = "00001234-5678-90ab-feDc-ba0987600000";
```

stevenhartley (approved):
LGTM

src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,122 @@
...
+	   << "-" << std::setw(4)
+	   << (((var & 0x3) << 14) |
+	       ((rand_b >> 48) & 0x3FFF))  // Variant and randb
+	   << "-" << std::setw(12) << (rand_b & 0xFFFFFFFFFFFF);  // Random number
+
+	return std::move(ss).str();
+}
+
+uprotocol::v1::UUID AsString::deserialize(const std::string& str) {
+	// Check if the UUID string is in the correct format
```
billpittman, line 49: NITPICK: How about a comment showing the "correct format"

// format : 12345678-1234-1234-1234-1234567890123
// index  : 012345678901234567890123456789012345

That would look better in monospace, but I can't figure out how to do it in a comment.

src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,122 @@
...
+
+namespace uprotocol::datamodel::serializer::uuid {
+
+std::string AsString::serialize(const uprotocol::v1::UUID uuid) {
+	// Extracting the parts of the UUIDv8
+	uint64_t unix_ts_ms = (uuid.msb() >> 16) & 0xFFFFFFFFFFFF;  // 48 bit
+	uint8_t ver = (uuid.msb() >> 12) & 0xF;                     // 4 bit
+	uint16_t counter = uuid.msb() & 0xFFF;                      // 12 bit
+	uint8_t var = (uuid.lsb() >> 62) & 0x3;                     // 2 bit
+	uint64_t rand_b = uuid.lsb() & 0x3FFFFFFFFFFFFFFF;          // 62 bit
```
billpittman, line 28: Lots of shifting and bitmasking -- maybe more constexpr would help make it clearer ?  The UUID validator cleaned up amazing well, I'm hoping for the same(ish).

src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,122 @@
...
+		ver = (ver_counter >> 12) & 0xF;
+		counter = ver_counter & 0xFFF;
+
+		uint16_t var_randb = std::stoul(str.substr(19, 4), nullptr, 16);
+		var = (var_randb >> 14) & 0x3;
+		rand_b = static_cast<uint64_t>(var_randb & 0x3FFF) << 48;
+		rand_b |= std::stoull(str.substr(24), nullptr, 16);
+	} catch (const std::exception& e) {
+		throw std::invalid_argument("Invalid UUID string format");
+	}
```
billpittman, line 77: NITPICK: Sorry :(  -- again maybe a comment showing how the UUID is deserialized

// format : 12345678-1234-1234-1234-1234567890123
// layout : __msb___-_msb-vcnt-____-_____RAND___
// v - version
// cnt -counter

OR whatever - I know you know this stuff now, but in 6 months we're ALL going to forget it.

src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,122 @@
...
+	uint32_t msb_low = htonl(static_cast<uint32_t>(uuid.msb() & 0xFFFFFFFF));
+	uint32_t lsb_high = htonl(static_cast<uint32_t>(uuid.lsb() >> 32));
+	uint32_t lsb_low = htonl(static_cast<uint32_t>(uuid.lsb() & 0xFFFFFFFF));
+
+	std::memcpy(bytes.data(), &msb_high, sizeof(uint32_t));
+	std::memcpy(bytes.data() + sizeof(uint32_t), &msb_low, sizeof(uint32_t));
+	std::memcpy(bytes.data() + 2 * sizeof(uint32_t), &lsb_high,
+	            sizeof(uint32_t));
+	std::memcpy(bytes.data() + 3 * sizeof(uint32_t), &lsb_low,
+	            sizeof(uint32_t));
```
billpittman, line 100: I know what you're doing, but this is a bit unclear.  Is there no bytes.data[4] way of indexing this array ?
I know these are 32-bit values being manipulated, so fixed offsets 0, 4, 8, 12 wouldn't be confusing to anyone.
I might even go so far as 
constexpr int MSB_HIGH=0
constexpr int MSB_LOW_ =4    (extra underscore to make the constants the same length so the code lines up better?)

I'm less firm about this change, so use your best judgement.

src/datamodel/serializer/Uuid.cpp
```diff
@@ -10,3 +10,122 @@
...
+	std::memcpy(&msb_low, bytes.data() + sizeof(uint32_t), sizeof(uint32_t));
+	std::memcpy(&lsb_high, bytes.data() + 2 * sizeof(uint32_t),
+	            sizeof(uint32_t));
+	std::memcpy(&lsb_low, bytes.data() + 3 * sizeof(uint32_t),
+	            sizeof(uint32_t));
+
+	uint64_t msb =
+	    (static_cast<uint64_t>(ntohl(msb_high)) << 32) | ntohl(msb_low);
+	uint64_t lsb =
+	    (static_cast<uint64_t>(ntohl(lsb_high)) << 32) | ntohl(lsb_low);
```
billpittman, line 122: See serialize comments.

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,235 @@ class TestFixture : public testing::Test {
...
+// Test string serialization
+TEST_F(TestUuidSerializer, SerializeToString) {
+	uprotocol::v1::UUID uuid;
+	uuid.set_msb(0x1234567890ABCDEF);
+	uuid.set_lsb(0xFEDCBA0987654321);
+
+	std::string serialized_uuid =
+	    uprotocol::datamodel::serializer::uuid::AsString::serialize(uuid);
+	// Assert the serialized UUID matches the expected value
+	EXPECT_EQ(serialized_uuid, "12345678-90ab-cdef-fedc-ba0987654321");
```
billpittman, line 48: Makes me want to ask --- The code doesn't care about case .... does the spec specify lower vs upper case for hex digits?

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,235 @@ class TestFixture : public testing::Test {
...
+	    uprotocol::datamodel::serializer::uuid::AsString::serialize(uuid);
+	// Assert the serialized UUID matches the expected value
+	EXPECT_EQ(serialized_uuid, "00001234-0078-00ab-00fe-00ba09876543");
+}
+
+// Test serialization with mixed case letters
+TEST_F(TestUuidSerializer, SerializeWithMixedCaseLetters) {
+	uprotocol::v1::UUID uuid;
+	uuid.set_msb(0x1234567890ABcDEF);
+	uuid.set_lsb(0x00dcbA0987654321);
```
billpittman, line 67: KUDOS : I don't know if you intentionally mixed the cases here - but GOOD test!

billpittman: Oh you did do it on purpose (just saw the name of the test) -- well comment still applies - good test!

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,235 @@ class TestFixture : public testing::Test {
...
+	// Assert that deserialization throws an invalid argument exception
+	EXPECT_THROW(uprotocol::datamodel::serializer::uuid::AsString::deserialize(
+	                 invalid_uuid_str),
+	             std::invalid_argument);
+}
+
+// Test deserialization with correct length but incorrect placement of dashes
+TEST(DeserializerTest, DeserializeWithMissingOneCharacter) {
+	std::string invalid_uuid =
+	    "12345678-1234-5678-1234-56781234567";  // Missing one character
```
billpittman, line 138: Ok there's a missing one character test .... is there an extra character test ?  (IE too long of a string).  If there is, I must have missed it.

billpittman (changes requested):
Mostly nitpicks - use your best judgement on those.

On the tests, looks great, I liked your use of sequential (expected, easy-to-identify) strings.

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,235 @@ class TestFixture : public testing::Test {
...
+}
+
+// Test invalid string deserialization
+TEST(DeserializerTest, InvalidUUIDFormat) {
+	// Define an invalid UUID string (missing dashes)
+	std::string invalid_uuid_str = "123456789abcdef0123456789abcdef0";
+	// Assert that deserialization throws an invalid argument exception
+	EXPECT_THROW(uprotocol::datamodel::serializer::uuid::AsString::deserialize(
+	                 invalid_uuid_str),
+	             std::invalid_argument);
```
billpittman, line 132: Hmmm - for the purpose of code coverage, shouldn't this be multiple tests, one for each dash position ?
x----
-x---
--x-- etc ?

test/coverage/datamodel/UuidSerializerTest.cpp
```diff
@@ -23,16 +27,235 @@ class TestFixture : public testing::Test {
...
+	                 invalid_uuid),
+	             std::invalid_argument);
+}
+
+TEST(DeserializerTest, DeserializeWithIncorrectDashPlacement) {
+	std::string invalid_uuid =
+	    "123456781-2345-6781-2345-67812345678";  // Dash placement incorrect
+	EXPECT_THROW(uprotocol::datamodel::serializer::uuid::AsString::deserialize(
+	                 invalid_uuid),
+	             std::invalid_argument);
```
billpittman, line 149: Actually maybe this is the one that should be split into multiple tests -- totally up to you.

test/coverage/datamodel/UuidSerializerTest.cpp
BishrutSubedi: there are some other cases of mixed cases which should cover this

test/coverage/datamodel/UuidSerializerTest.cpp
BishrutSubedi: In term of coverage this should be good, but I'll add few more test for other dash placement

test/coverage/datamodel/UuidSerializerTest.cpp
BishrutSubedi: yeah, this test case slipped my mind. I'll add it in next commit

BishrutSubedi:
Thank you, I've updated PR with the suggested changes

billpittman:
> Thank you, I've updated PR with the suggested changes

No thank you.  Looks great - thanks for the extra effort.  Approved!

billpittman (approved):

src/datamodel/serializer/Uuid.cpp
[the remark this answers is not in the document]
gregmedd: Since we got a LGTM from @stevenhartley, I'm going to go ahead and resolve this.

test/coverage/datamodel/UuidSerializerTest.cpp
gregmedd: The spec is case insensitive: https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-format

gregmedd (approved):
Looks good to me. Can you make sure that the commits are rebased so that the conan change is dropped? It should already be committed to main. Otherwise this is ready to go.

PR #140, opened 2024-06-04
UUri: Validator first check-in

billpittman:
First check-in for UUri validator.

### DIFF
--- a/include/up-cpp/datamodel/validator/UUri.h
+++ b/include/up-cpp/datamodel/validator/UUri.h
@@ -110,7 +110,7 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 ///   * The uE ID is 0
 ///   * The uE major version is 0
 ///   * The resource ID is 0
-[[nodiscard]] ValidationResult isEmpty(const v1::UUri& uri);
+[[nodiscard]] ValidationResult isEmpty(const v1::UUri&);
 
 /// @brief Checks if a UUri is local
 ///
--- a/include/up-cpp/datamodel/validator/UUri.h
+++ b/include/up-cpp/datamodel/validator/UUri.h
@@ -54,7 +54,6 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 /// A UUri is valid if:
 ///
 ///   * It is not empty
-///   * It is reserved (or one of the fields is reserved)
 ///   * it is valid for at least one of:
 ///     * isValidRpcMethod()
 ///     * isValidRpcResponse()
@@ -64,13 +63,13 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 
 /// @brief Checks if UUri is valid for invoking an RPC method
 ///
-/// The UUri must not be blank/reserved, no field can be a wildcard, and
+/// The UUri must not be blank, no field can be a wildcard, and
 /// resource_id must be in the range [0x0001, 0x7FFF].
 [[nodiscard]] ValidationResult isValidRpcMethod(const v1::UUri&);
 
 /// @brief Checks if UUri is a valid sink for responding to an RPC request.
 ///
-/// The UUri must not be blank/reserved, no field can be a wildcard, and
+/// The UUri must not be blank, no field can be a wildcard, and
 /// resource_id must be 0.
 [[nodiscard]] ValidationResult isValidRpcResponse(const v1::UUri&);
 
@@ -85,14 +84,20 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 ///        and sink for sending notifications, OR as a sink for receiving
 ///        notifications.
 ///
-/// The UUri must not be blank/reserved, no field can be a wildcard, and
+/// The UUri must not be blank, no field can be a wildcard, and
 /// resource_id must be in the range [0x8000, 0xFFFE].
-[[nodiscard]] ValidationResult isValidTopic(const v1::UUri&);
+[[nodiscard]] ValidationResult isValidPublishTopic(const v1::UUri&);
+
+/// @brief Checks if UUri is valid for notification message.
+///
+/// The UUri must not be blank, no field can be a wildcard, and
+/// resource_id must be in the range [0x8000, 0xFFFE].
+[[nodiscard]] ValidationResult isValidNotification(const v1::UUri&);
 
 /// @brief Checks if UUri is valid as a subscription to a published topic or
 ///        as a source filter when subscribing to a notification.
 ///
-/// The UUri must not be blank/reserved, and resource_id, if not a wildcard,
+/// The UUri must not be blank, and resource_id, if not a wildcard,
 /// must be in the range [0x8000, 0xFFFE].
 [[nodiscard]] ValidationResult isValidSubscription(const v1::UUri&);
 
@@ -107,19 +112,15 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 ///   * The resource ID is 0
 [[nodiscard]] ValidationResult isEmpty(const v1::UUri& uri);
 
-/// @brief Checks if a URI contains an *always* reserved value.
-///
-/// @note uE Service ID 0 is reserved *for uSubscription*, and is not reserved
-///       in all cases. As such, it is not checked for here.
-///
-/// A UUri is considered reserved if either uE major version is 0 or resource
-/// ID is 0.
-[[nodiscard]] ValidationResult isReserved(const v1::UUri& uri);
-
 /// @brief Checks if a UUri is local
 ///
 /// This is just a check for a zero-length authority name string.
-[[nodiscard]] bool isLocal(const v1::UUri& uri);
+[[nodiscard]] bool isLocal(const v1::UUri&);
+
+/// @brief  Checks if a UUri uses wildcards
+///
+/// Checks for all types of wildcards, returns true if any are found.
+[[nodiscard]] bool uses_wildcards(const v1::UUri&);
 
 /// @brief This exception indicates that a UUri object was provided that
 ///        did not contain valid UUri data.
--- a/src/datamodel/validator/UUri.cpp
+++ b/src/datamodel/validator/UUri.cpp
@@ -10,3 +10,169 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "up-cpp/datamodel/validator/UUri.h"
+
+namespace uprotocol::datamodel::validator::uri {
+
+using namespace uprotocol;
+
+std::string_view message(Reason reason) {
+    switch (reason) {
+	case Reason::EMPTY:
+	    return "The URI authority_name is empty or whitespace.";
+	case Reason::RESERVED_VERSION:
+	    return "The version is 0 (reserved).";
+	case Reason::RESERVED_RESOURCE:
+	    return "The resource is 0 for non-RPC-response URIs.";
+	case Reason::DISALLOWED_WILDCARD:
+	    return "The URI contains a wildcard in one or more fields that is not allowed for this form.";
+        case Reason::BAD_RESOURCE_ID:
+	    return "The resource ID is not in the allowed range for this URI form.";
+	default:
+	    return "Unknown reason.";
+    }
+}
+
+bool uses_wildcards(const v1::UUri& uuri) {
+    if (uuri.authority_name().find("*") != std::string::npos) {
+        return true;
+    }
+    if ((uuri.ue_id() & 0xFFFF) == 0xFFFF) { // service ID
+        return true;
+    }
+    if ((uuri.ue_id() & 0xFFFF0000) == 0) {  // service instance ID
+        return true;
+    }
+    if (uuri.ue_version_major() == 0xFF) {
+        return true;
+    }
+    if (uuri.resource_id() == 0xFFFF) {
+        return true;
+    }
+    return false;
+}
+
+ValidationResult isValid(const v1::UUri& uuri) {
+    auto [valid, reason] = isValidRpcMethod(uuri);
+    if (!valid) {
+        auto [valid, reason] = isValidRpcResponse(uuri);
+        if (!valid) {
+            auto [valid, reason] = isValidPublishTopic(uuri);
+            if (!valid) {
+                return isValidNotification(uuri);
+            }
+        }
+    }
+
+    return { true, std::nullopt };
+}
+
+ValidationResult isValidRpcMethod(const v1::UUri& uuri) {
+    if (uuri.authority_name().empty()) {
+        return { false, Reason::EMPTY };
+    }
+
+    // disallow wildcards
+    if (uses_wildcards(uuri)) {
+        return { false, Reason::DISALLOWED_WILDCARD };
+    }
+
+    // check resource ID [0x0001, 0x7FFF]
+    uint16_t resource_id = uuri.resource_id() & 0xFFFF;
+    if (resource_id == 0 || resource_id > 0x7FFF) {
+        return { false, Reason::BAD_RESOURCE_ID };
+    }
+
+    return { true, std::nullopt };
+}
+
+ValidationResult isValidRpcResponse(const v1::UUri& uuri) {
+    if (uuri.authority_name().empty()) {
+        return { false, Reason::EMPTY };
+    }
+
+    // disallow wildcards
+    if (uses_wildcards(uuri)) {
+        return { false, Reason::DISALLOWED_WILDCARD };
+    }
+
+    if (uuri.resource_id() != 0) {
+        return { false, Reason::BAD_RESOURCE_ID };
+    }
+    return { true, std::nullopt };
+}
+
+ValidationResult isValidDefaultSource(const v1::UUri& uuri) {
+    return isValidRpcResponse(uuri);
+}
+
+ValidationResult isValidPublishTopic(const v1::UUri& uuri) {
+    if (uuri.authority_name().empty()) {
+        return { false, Reason::EMPTY };
+    }
+
+    // disallow wildcards
+    if (uses_wildcards(uuri)) {
+        return { false, Reason::DISALLOWED_WILDCARD };
+    }
+
+    if ((uuri.resource_id() & 0xFFFF) < 0x8000) {
+        return { false, Reason::BAD_RESOURCE_ID };
+    }
+
+    return { true, std::nullopt };
+}
+
+ValidationResult isValidNotification(const v1::UUri& uuri) {
+    if (uuri.authority_name().empty()) {
+        return { false, Reason::EMPTY };
+    }
+
+    // disallow wildcards
+    if (uses_wildcards(uuri)) {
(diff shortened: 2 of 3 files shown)


src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,169 @@
...
+
+	if ((uuri.resource_id() & 0xFFFF) < 0x8000) {
+		return {false, Reason::BAD_RESOURCE_ID};
+	}
+
+	return isValidPublishTopic(uuri);
+}
+
+ValidationResult isEmpty(const v1::UUri& uuri) {
+	if (uuri.authority_name().find_first_not_of(" ") != std::string::npos) {
```
gregmedd, line 157: There might be other whitespace characters we should look for. Digging into the [history of this repo](https://github.com/eclipse-uprotocol/up-cpp/blob/up-v1.5.7/include/up-cpp/uri/tools/Utils.h#L51), this check used to be done using [`std::isspace()`](https://en.cppreference.com/w/cpp/locale/isspace). There's also [`std::isblank()`](https://en.cppreference.com/w/cpp/locale/isblank), which _might_ be a better choice (I'm actually not entirely clear the difference).

test/coverage/datamodel/UUriValidatorTest.cpp
```diff
@@ -12,9 +12,13 @@
 #include <gtest/gtest.h>
 #include <up-cpp/datamodel/validator/UUri.h>
 
+#define AUTHORITY_NAME "test"
```
gregmedd, line 15: I recommend moving this into the anonymous namespace and switching to a constexpr.

```suggestion
constexpr std::string_view AUTHORITY_NAME{"test"};
```

test/coverage/datamodel/UUriValidatorTest.cpp
```diff
@@ -23,16 +27,296 @@ class TestFixture : public testing::Test {
...
+	uuri.set_authority_name(AUTHORITY_NAME);
+	uuri.set_ue_id(0x00010001);
+	uuri.set_ue_version_major(1);
+	uuri.set_resource_id(1);
+
+	uuri.set_authority_name("");
+	auto res = isValid(uuri);
+	EXPECT_FALSE(std::get<bool>(res));
+	EXPECT_TRUE(std::get<std::optional<Reason>>(res) == Reason::EMPTY);
+	uuri.set_authority_name(AUTHORITY_NAME);
```
gregmedd, line 50: I would recommend structuring a test like this such that a new uuri is created for each configuration that is checked.  The pattern I typically use looks like this:
```suggestion
    auto getUuri = []() {
	    uprotocol::v1::UUri uuri;
	    uuri.set_authority_name(AUTHORITY_NAME);
	    uuri.set_ue_id(0x00010001);
	    uuri.set_ue_version_major(1);
	    uuri.set_resource_id(1);
	    return uuri;
	};

	{
	    auto uuri = getUuri();
	    auto [valid, maybeReason] = isValid(uuri);
	    EXPECT_TRUE(valid);
	    EXPECT_FALSE(maybeReason);
	}

    {
	    auto uuri = getUuri();
	    uuri.set_authority_name("");
	    auto [valid, maybeReason] = isValid(uuri);
	    EXPECT_FALSE(valid);
	    EXPECT_TRUE(maybeReason);
	    EXPECT_EQ(*maybeReason, Reason::EMPTY);
	}
```

gregmedd (changes requested):
A few edits, along with some questions for @stevenhartley.

Overall it looks pretty good so you should be able to get it committed once those items are resolved.

include/up-cpp/datamodel/validator/UUri.h
```diff
@@ -64,13 +63,13 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 
 /// @brief Checks if UUri is valid for invoking an RPC method
 ///
-/// The UUri must not be blank/reserved, no field can be a wildcard, and
+/// The UUri must not be blank, no field can be a wildcard, and
 /// resource_id must be in the range [0x0001, 0x7FFF].
```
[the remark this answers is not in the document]
stevenhartley: Range is specific to the use of UUri (sink/source per message type) so they are now documented in [uattributes.adoc](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc)

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,169 @@
...
+using namespace uprotocol;
+
+std::string_view message(Reason reason) {
+	switch (reason) {
+		case Reason::EMPTY:
+			return "The URI authority_name is empty or whitespace.";
+		case Reason::RESERVED_VERSION:
+			return "The version is 0 (reserved).";
+		case Reason::RESERVED_RESOURCE:
+			return "The resource is 0 for non-RPC-response URIs.";
```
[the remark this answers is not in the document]
stevenhartley: See previous comment, they have just moved to attributes

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,169 @@
...
+			       "form.";
+		default:
+			return "Unknown reason.";
+	}
+}
+
+bool uses_wildcards(const v1::UUri& uuri) {
+	if (uuri.authority_name().find("*") != std::string::npos) {
+		return true;
+	}
```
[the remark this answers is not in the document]
stevenhartley: hum, good point. Not specified yet.

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,169 @@
...
+		return {false, Reason::EMPTY};
+	}
+
+	// disallow wildcards
+	if (uses_wildcards(uuri)) {
+		return {false, Reason::DISALLOWED_WILDCARD};
+	}
+
+	// check resource ID [0x0001, 0x7FFF]
+	uint16_t resource_id = uuri.resource_id() & 0xFFFF;
```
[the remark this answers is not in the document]
stevenhartley: It is limited to 16 bits but proto only has uint32 size.

stevenhartley (commented):
see feedback

src/datamodel/validator/UUri.cpp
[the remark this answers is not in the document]
billpittman: I ran into this too wrt the spec, but I figured a wildcard is a wildcard even if its "partially" expressed.  For example  "127.0.0.*" still uses a wildcard and shouldn't be allowed in the validator.

include/up-cpp/datamodel/validator/UUri.h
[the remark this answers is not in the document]
gregmedd: Ah! That makes sense. Thanks!

src/datamodel/validator/UUri.cpp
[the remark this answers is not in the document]
gregmedd: @billpittman we might need to modify this check to enforce the range without bitmasking in case the upper 16 bits get set incorrectly.

gregmedd:
BTW, you will probably have to rebase onto main from the eclipse repo to get the CI fix.

include/up-cpp/datamodel/validator/UUri.h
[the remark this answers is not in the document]
billpittman: I'm assuming this is closed since we found the reference in the spec.  Reopen if I'm wrong.

include/up-cpp/datamodel/validator/UUri.h
```diff
@@ -107,19 +112,15 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
...
-///
-/// A UUri is considered reserved if either uE major version is 0 or resource
-/// ID is 0.
-[[nodiscard]] ValidationResult isReserved(const v1::UUri& uri);
-
 /// @brief Checks if a UUri is local
 ///
 /// This is just a check for a zero-length authority name string.
-[[nodiscard]] bool isLocal(const v1::UUri& uri);
+[[nodiscard]] bool isLocal(const v1::UUri&);
```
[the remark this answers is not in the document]
billpittman: Ah missed one.  Thanks!

src/datamodel/validator/UUri.cpp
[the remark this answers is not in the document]
billpittman: I'm going to close this since the code catches both "*" or "foo*".

src/datamodel/validator/UUri.cpp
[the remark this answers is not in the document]
billpittman: @gregmedd Are you trying to say that you want me to verify that the upper 16 bits are correctly set to 0 ?

src/datamodel/validator/UUri.cpp
billpittman: I rewrote it to use isspace (catches all space type characters, tab, newline, carriage return, etc).

test/coverage/datamodel/UUriValidatorTest.cpp
billpittman: Not exactly what you asked for wrt std::string_view, but similar.  The template functions really hated me using std::string_view and threw a bunch of compiler warnings "template argument deduction/substitution failed".  This change accomplishes what you were looking for I believe.

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,169 @@
...
+		if (!valid) {
+			auto [valid, reason] = isValidPublishTopic(uuri);
+			if (!valid) {
+				return isValidNotification(uuri);
+			}
+		}
+	}
+
+	return {true, std::nullopt};
+}
```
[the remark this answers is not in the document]
billpittman: I'm FAR less pleased with this implementation, but if you're willing to accept it, then I'll live with it.

test/coverage/datamodel/UUriValidatorTest.cpp
billpittman: Done.  I do have to admit, I was originally against this change and felt it was unnecessary, however after doing the change, I can honestly say I like it a LOT BETTER than before.

src/datamodel/validator/UUri.cpp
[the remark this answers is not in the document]
billpittman: I changed the code to expect the upper 16-bits to be 0 and removed the bitmasking.

billpittman:
> BTW, you will probably have to rebase onto main from the eclipse repo to get the CI fix.

FWIW, I didn't have to rebase, simply changing my branch was enough.  Github essentially did the rebase for me since it pulled my change onto main to perform the CI.

BishrutSubedi (approved):
LGTM

include/up-cpp/datamodel/validator/UUri.h
```diff
@@ -64,13 +63,13 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 
 /// @brief Checks if UUri is valid for invoking an RPC method
 ///
-/// The UUri must not be blank/reserved, no field can be a wildcard, and
```
gregmedd, line 67: Reading between the lines in the spec, reserved values are still a thing. The comments in this header should be restored.

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,179 @@
...
+	if (uuri.authority_name().empty()) {
+		return {false, Reason::EMPTY};
+	}
+
+	// disallow wildcards
+	if (uses_wildcards(uuri)) {
+		return {false, Reason::DISALLOWED_WILDCARD};
+	}
+
+	if ((uuri.resource_id() & 0xFFFF) < 0x8000) {
```
gregmedd, line 129: This also needs to avoid bitmasking:

```suggestion
	if (uuri.resource_id() < 0x8000 || uuri.resource_id() > 0x7FFF) {
```

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,179 @@
...
+		return {false, Reason::BAD_RESOURCE_ID};
+	}
+
+	return {true, std::nullopt};
```
gregmedd, line 150: I might be wrong, but this looks like it might just be a copy of `isValidPublishTopic()`

```suggestion
	return isValidPublishTopic();
```

Although I thought that they were originally one combined check :thinking:

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,179 @@
...
+}
+
+ValidationResult isValidSubscription(const v1::UUri& uuri) {
+	if (uuri.authority_name().empty()) {
+		return {false, Reason::EMPTY};
+	}
+
+	if ((uuri.resource_id() & 0xFFFF) < 0x8000) {
```
gregmedd, line 158: See comment on bit masking

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,179 @@
...
+}
+
+ValidationResult isEmpty(const v1::UUri& uuri) {
+	if (!std::all_of(uuri.authority_name().begin(), uuri.authority_name().end(),
+	                 isspace)) {
+		return {false, Reason::EMPTY};
+	}
+
+	if (uuri.ue_id() != 0) {
+		return {false, Reason::RESERVED_RESOURCE};
```
gregmedd, line 172: I just noticed that `RESERVED_RESOURCE` is being used for a non-zero `ue_id`. Incidentally, I think 0 is allowed for both `ue_id` and `resource_id` under some circumstances.

test/coverage/datamodel/UUriValidatorTest.cpp
gregmedd: Thanks for making the change. It's a lot easier to read what the test is doing now.

src/datamodel/validator/UUri.cpp
gregmedd: TBH this is really minor. I'm not sure if I would stop the review over it.

gregmedd (commented):
Just a couple more small changes and this should be ready to merge.

src/datamodel/validator/UUri.cpp
billpittman: You're right, it is a copy, technically.  According to the spec, a notification message could be a source or a sink, so I made the assumption that a source would be more likely to check for "being valid" (ie check before I send this).  So the test is for a source (since a UUri doesn't specify within itself whether its a source or a sink).

I'd like to keep the separate, because I'm guessing its going to change, or even split into isValidNotificationSource and isValidNotificationSink (or similar).

billpittman: Thinking more about it, I'm going to change the validator code to also approve "==0" since technically that's also valid.  Maybe that's what past-Bill was thinking, but he must have forgotten somehow.

src/datamodel/validator/UUri.cpp
billpittman: Good, because that check is part of isEmpty() and checking for an empty UUri that has a non-zero ue_id seems right - although we could argue that the error code is wrong.  But that one seems "fitting" unless I wanted to make a new one.

For the record - I'm leaving it unchanged.

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,179 @@
...
+}
+
+bool uses_wildcards(const v1::UUri& uuri) {
+	if (uuri.authority_name().find("*") != std::string::npos) {
+		return true;
+	}
+	if ((uuri.ue_id() & 0xFFFF) == 0xFFFF) {  // service ID
+		return true;
+	}
+	if ((uuri.ue_id() & 0xFFFF0000) == 0) {  // service instance ID
```
ruchirchauhan, line 44: @stevenhartley, correct me if am wrong, but Instance Id can be 0. As per someip (pr in review) spec if instance id is 0 it is treated as 1 and can not be treated as a wildcard entry.

ruchirchauhan: This is from a PR, not merged yet in mainline, and is part of someip spec
https://github.com/int0x27/up-spec/blob/someip-spec-update/up-l1/someip.adoc#21-uuris

gregmedd: Make sure to verify ranges in the [uattributes spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc). I believe 0 is excluded from most ranges in the URI, with the exception of resource where 0 is the destination for RPC responses

gregmedd, line 44: Isn't this an `and` condition - the ID must be 0xFFFF _and_ the instance ID must be 0?

src/datamodel/validator/UUri.cpp
```diff
@@ -10,3 +10,179 @@
...
+	if (uuri.authority_name().empty()) {
+		return {false, Reason::EMPTY};
+	}
+
+	// disallow wildcards
+	if (uses_wildcards(uuri)) {
+		return {false, Reason::DISALLOWED_WILDCARD};
+	}
+
+	if ((uuri.resource_id() < 0x8000) && (uuri.resource_id() != 0)) {
```
gregmedd, line 129: Publish can't use resource ID 0. That's reserved for RPC responses only

gregmedd: Also, doesn't this still have to check if the resource ID is above 0xFFFF? Because the protobuf struct uses uint32, but the resource ID is only allowed to be a uint16.

include/up-cpp/datamodel/validator/UUri.h
gregmedd: To clarify, if the ranges are combined, there are still some values that are reserved or disallowed. However, the new spec is incredibly vague and provides none of the guidance that was available before. I think we may need to file a bug to get the spec cleaned up then we can revisit.

src/datamodel/validator/UUri.cpp
billpittman: I don't think so, the basics/uri.adoc section 5 "Pattern Matching" has 5 bullet points, and nothing says they depend on each other.

Also in section 5.1 there's a couple examples where the upper and lower halves behave independently.

src/datamodel/validator/UUri.cpp
billpittman: As to the first --- right change, wrong function.  I intended for that change to go into isValidNotification, but made an error.  Same code was moved to its intended location now.

billpittman: As for the > 0xFFFF, let me check the whole file for that sort of thing, and I think adding a test would be good too.

src/datamodel/validator/UUri.cpp
gregmedd: It sounds like we need some clarity in the UURI specs. Let's get that first and we can revisit.

gregmedd: Ok, good enough for me :shrug:

gregmedd (approved):
LGTM

PR #142, opened 2024-06-05
clang-format: Extend default line length

billpittman:
Line length of 80 characters is too short in several cases where nested statements are used and long variable names cause the line to be wrapped and makes the code less readable.  Additionally, most tools (vscode, etc) allow for showing far more than 80 characters per line.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -11,6 +11,7 @@ DerivePointerAlignment: 'false'
 ExperimentalAutoDetectBinPacking: 'false'
 FixNamespaceComments: 'true'
 PointerAlignment: Left
+ColumnLimit: 100
 
 TabWidth: 4
 IndentWidth: 4


gregmedd:
We are not making format changes until a publicly available style guide has been published and approved.

billpittman:
> We are not making format changes until a publicly available style guide has been published and approved.

Please treat this as merely a suggestion/proposal that matches my preferences.

gregmedd:
Personally, I often end up working in windows that are <100 columns wide, so I tend to prefer an 80 column limit.

When we work on the style guide, I would be open to reviewing specific code examples where 80 columns is making the code harder to read, alongside comparisons of other column width limits for the same code.

PLeVasseur:
We decided to take a different path here, so closing

PR #144, opened 2024-06-05
Rolling back to Conan 2.3.2

gregmedd:
Conan 3.4.0 was released today, and it fixed some bugs that we were using in our recipes. See https://github.com/conan-io/conan/pull/16272

Additionally, the old way of passing the conan cache was brittle, so it has been switched to an artifact.

Lastly, a fixed conan profile is set to reduce the risk of changes breaking things.

### DIFF
--- a/.github/workflows/ci_conan_profile
+++ b/.github/workflows/ci_conan_profile
@@ -1,0 +1,8 @@
+[settings]
+arch=x86_64
+build_type=Release
+compiler=gcc
+compiler.cppstd=gnu17
+compiler.libcxx=libstdc++11
+compiler.version=11
+os=Linux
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_conan_profile
stevenhartley, line 1: Why do we need this now and not before? Just curious

gregmedd: Technically we don't, but I already had 99% of this fix on another branch, so I pulled that. It included this change.

What it brings with is us more resilience against unexpected environment changes. The Conan documentation warns that the output of `conan profile detect` is not stable and should not be used in CI environments.

stevenhartley (approved):
LGTM

PR #147, opened 2024-06-08
DRAFT: Add callbacks for RpcClient::invokeMethod

gregmedd:
Looking at replacing futures / adding callbacks to RpcClient. Different architectures may benefit from each. We could potentially implement the futures version in terms of the callback version.


gregmedd:
Superseded by #155 - this one has its source branch in the wrong repo.

Issue #149, opened 2024-06-10
Uri Validator IsValid API () is not allowing to have URI object with empty authority


pranavishere2:
**Code snippet:**
    v1::UUri uri;
    uri.set_authority_name("");
    uri.set_ue_id(0x10010001);
    uri.set_ue_version_major(0xFE);
    uri.set_resource_id(0x7500);
    **auto [valid, reason] = isValid(uri);**

**Expected Result:**
It should result valid as true. Authority can be empty as per the spec https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uri.adoc. Refer section 3.1.Examples

**Actual Result:**
It returns valid as false and reason as Reason::EMPTY

billpittman:
Yea I'll take a look.

PR #150, opened 2024-06-10
UUriValidator : authority_name can be empty (sometimes)

billpittman:
Allow authority_name to be empty in most cases, but specifically should not be empty in isValidDefaultSource.  Also updated the tests to test an empty authority_name and pass except when isValidDefaultSource is used.

Addresses issue #149


gregmedd (approved):
LGTM

pranavishere2 (approved):

PR #151, opened 2024-06-11
This PR adds UUID builder for uprotocol uuidv8

BishrutSubedi:
Add uuid builder and it's unit test

TODO:

Convert from uuidv8 to uuidv7
Follow up on thread safety


src/datamodel/builder/Uuid.cpp
```diff
@@ -10,3 +10,99 @@
...
+	uuid.set_msb(msb);
+	uuid.set_lsb(lsb);
+	return uuid;
+}
+
+UuidBuilder::UuidBuilder(bool testing)
+    : testing_(testing),
+      time_source_(nullptr),
+      random_source_(nullptr),
+      shared_state_(std::make_shared<UuidSharedState>()) {}
```
gregmedd, line 106: There should be a static shared state that is shared across all UuidBuilder instances so that the counters are synchronized. You could add a static function to UuidBuilder that looks something like this:

```
static std::once_flag init_shared_state_once_;
static std::shared_ptr<UuidSharedState> default_state_;
static std::shared_ptr<UuidSharedState> getDefaultState() {
    std::call_once(init_shared_state_once_, [&default_state_](){
        default_state_ = std::make_shared<UuidSharedState>();
    });
    return default_state_;
}
```

src/datamodel/builder/Uuid.cpp
```diff
@@ -10,3 +10,99 @@
...
+#include <stdexcept>
+
+namespace uprotocol::datamodel::builder {
+
+struct UuidBuilder::UuidSharedState {
+	std::mt19937_64 random_engine{std::random_device{}()};
+	uint16_t counter{0};
+	uint64_t last_unix_ts_ms{0};
+	uint64_t rand_b{0};
+	bool rand_b_initialized{false};
```
gregmedd, line 24: These values may need to be wrapped in `std::atomic` since this state will be shared across UuidBuilder instances

src/datamodel/builder/Uuid.cpp
```diff
@@ -10,3 +10,99 @@
...
 
 #include "up-cpp/datamodel/builder/Uuid.h"
+
+#include <random>
+#include <stdexcept>
+
+namespace uprotocol::datamodel::builder {
+
+struct UuidBuilder::UuidSharedState {
+	std::mt19937_64 random_engine{std::random_device{}()};
```
gregmedd, line 20: Is this thread safe? Maybe the best thing to do is add a constructor to this struct that generates rand_b then this member and the `rand_b_initialized` member can be removed.

Making that change along with the `std::call_once` in another comment would make the initialization of rand_b safe.

src/datamodel/builder/Uuid.cpp
```diff
@@ -10,3 +10,99 @@
...
+		// Reset the counter if the timestamp tick has changed
+		shared_state_->counter = 0;
+		shared_state_->last_unix_ts_ms = unix_ts_ms;
+	}
+
+	uint64_t msb = static_cast<uint64_t>(unix_ts_ms) << 16;
+
+	msb |= static_cast<uint64_t>(8) << 12;  // Set the version to 8
+
+	if (shared_state_->counter == 4095) {
```
gregmedd, line 77: I'm left wondering if some of these constants, shifts, masks, and magic numbers couldn't be constexpr values. And if they could be shared with other UUID code. Maybe they could go in a `up-cpp/datamodel/common/UuidConstants.h` file

src/datamodel/builder/Uuid.cpp
```diff
@@ -10,3 +10,99 @@
...
+
+	uint64_t lsb;
+	if (!shared_state_->rand_b_initialized) {
+		shared_state_->rand_b = random_source_
+		                            ? random_source_()
+		                            : std::uniform_int_distribution<uint64_t>{}(
+		                                  shared_state_->random_engine) &
+		                                  0x3FFFFFFFFFFFFFFF;
+		shared_state_->rand_b_initialized = true;
+	}
```
gregmedd, line 92: See comment about handling the initialization in the constructor for the shared state struct

src/datamodel/builder/Uuid.cpp
```diff
@@ -10,3 +10,99 @@
...
+
+#include <random>
+#include <stdexcept>
+
+namespace uprotocol::datamodel::builder {
+
+struct UuidBuilder::UuidSharedState {
+	std::mt19937_64 random_engine{std::random_device{}()};
+	uint16_t counter{0};
+	uint64_t last_unix_ts_ms{0};
```
gregmedd, line 22: minor: could this be a `std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds>`? That way all the time comparisons and math are on chrono objects and the timestamp is only converted to an integer when it is ready to be inserted into the UUID

test/coverage/datamodel/UuidBuilderTest.cpp
```diff
@@ -10,29 +10,261 @@
...
+	auto builder = UuidBuilder::getTestBuilder().withRandomSource(
+	    [fixed_random]() { return fixed_random; });
+	auto uuid = builder.build();
+
+	EXPECT_EQ(uuid.lsb() & 0x3FFFFFFFFFFFFFFF,
+	          fixed_random & 0x3FFFFFFFFFFFFFFF);
+}
+
+// Test independent state
+TEST(UuidBuilderTest, WithIndependentState) {
```
gregmedd, line 73: Add a copy of this test for `WithDependentState` where you verify that two regular builders produce values with related states. For example, the `EXPECT_NE` on line 89 would turn into an `EXPECT_EQ`

test/coverage/datamodel/UuidBuilderTest.cpp
gregmedd, line 1: Add one test where multiple threads each hold a separate UuidBuilder instance and are generating UUIDs.

gregmedd (commented):
For the most part, this looks good to me. There are a few changes that could clean this up, and I think that the current implementation doesn't actually share state between instances.

stevenhartley:
> that could clean this up, and I think that the current

@gregmedd given the change now to move to UUIDv7 for the next release, we don't need to share state anymore as rand_a and rand_b is re-generated at each call to create()

src/datamodel/builder/Uuid.cpp
```diff
@@ -10,3 +10,99 @@
...
+	                      now.time_since_epoch())
+	                      .count();
+
+	if (unix_ts_ms != shared_state_->last_unix_ts_ms) {
+		// Reset the counter if the timestamp tick has changed
+		shared_state_->counter = 0;
+		shared_state_->last_unix_ts_ms = unix_ts_ms;
+	}
+
+	uint64_t msb = static_cast<uint64_t>(unix_ts_ms) << 16;
```
billpittman, line 73: This looks more like UUIDv7 instead of UUIDv8 (v7 says time, v8 says implementation specific).

src/datamodel/builder/Uuid.cpp
billpittman: In addition to what @gregmedd said above, it seems to me if the counter freezees, shouldn't that be some sort of error? If someone generates 5000 UUID's within one ms (possible I think), some of them would be identical ... right?

billpittman:
Looking at the code and RFC 9562 (referenced in the spec), this looks more like v7 instead of v8.  Is the title a typo ?

If this is supposed to be v8, it seems like all the variable names are aligning with the v7 field names instead of the v8 field names (ie unix_ts_ms vs custom_a, etc.).

src/datamodel/builder/Uuid.cpp
BishrutSubedi: Agree, the uuidv8 spec we have is very similar to v7 in term of bits arrangement

BishrutSubedi:
> Looking at the code and RFC 9562 (referenced in the spec), this looks more like v7 instead of v8. Is the title a typo ?
> 
> If this is supposed to be v8, it seems like all the variable names are aligning with the v7 field names instead of the v8 field names (ie unix_ts_ms vs custom_a, etc.).
Ah the spec were updated yesterday to change to uuidv7. For most part, I think the implementation should be same. I'll go over this to see if any specific changes are necessary

gregmedd:
On the change from v8 to v7, I lean towards proceeding with this PR as-is. I understand it is the wrong UUID version. We can take a follow-up task to correct the implementation back to uuid v7 and evaluate if any of the other modules need to change at the same time (e.g. the validator).

BishrutSubedi:
It looks like the counter field is replaced by randa. The remaining bits allocation are similar.

gregmedd:
Tracking change to UUIDv7 as #156

src/datamodel/builder/Uuid.cpp
gregmedd: Given the UUIDv7 change is coming, we can hold off on messing around with the shared state stuff. It can remain unshared for this PR.

src/datamodel/builder/Uuid.cpp
gregmedd: We can hold off on making the shared state thread safe for this PR

src/datamodel/builder/Uuid.cpp
gregmedd: When we change to UUIDv7, the rand field will update for every message, so they will still be uniquely identifiable (but possibly not sortable).

Then again, using UUID for absolute global ordering of message _has pitfalls_.

src/datamodel/builder/Uuid.cpp
gregmedd: Not worrying about shared state for this PR

test/coverage/datamodel/UuidBuilderTest.cpp
gregmedd: Not worrying about shared state for this PR

test/coverage/datamodel/UuidBuilderTest.cpp
gregmedd: Not worrying about shared state for this PR

src/datamodel/builder/Uuid.cpp
gregmedd: Not worrying about shared state for this PR

src/datamodel/builder/Uuid.cpp
BishrutSubedi: Thanks, I've updated this

src/datamodel/builder/Uuid.cpp
BishrutSubedi: I'll follow up with another PR to make this adhere with current v7 spec

BishrutSubedi:
>In addition to what @gregmedd said above, it seems to me if the counter freezees, shouldn't that be some sort of error? If someone generates 5000 UUID's within one ms (possible I think), some of them would be identical ... right?

I agree with this, but strictly going by the previous spec, counter is expected to freeze. This is a valid point though. I think throwing exception here can cause software calling build() function to crash, unless they check the counter value and handle it appropriately. May be returning `std:nullopt` would be better than raising exception. We're looking to change this to v7, but this is good discussion to have

gregmedd (approved):
LGTM. UUIDv7 will be implemented as a follow-up (#156)

PR #153, opened 2024-06-11
Pass UTransport's cleanup method for callbacks

### LINKED ISSUE #152: UTransport does not provide the cleanup function when registering callbacks
It needs to provide the cleanup function in order for it to be called when the `CalleeHandle` disconnects.

gregmedd:
This method should be called when the `CalleeHandle` breaks the connection so that derived transports can clean up connections should they choose to do so. Enable this by passing a lambda that calls the method.

closes #152


cbeck83 (approved):
The changes is generally clear and concise.

PR #155, opened 2024-06-12
Adding callbacks to RpcClient::invokeMethod

### LINKED ISSUE #108: Replace futures in RpcClient with callbacks
@sanjok-itin has pointed out that waiting on std::futures does not allow for a shutdown path. Either threads waiting on futures need to wake periodically to check if they need to stop waiting, or we need to wake them by breaking the promises.

We should look into implementing some sort of shutdown interface so that we can break promises and block any new calls to the transport APIs.

Alternatively, returning some sort of other waitable object that can be signaled from both ends of the connection could be used, or we could return something that can be waited on as a group (e.g. using `poll()` or `select()` or `epoll()`)

gregmedd:
Looking at replacing futures / adding callbacks to RpcClient. Different architectures may benefit from each. We could potentially implement the futures version in terms of the callback version.

closes #108

### DIFF
--- a/include/up-cpp/communication/RpcClient.h
+++ b/include/up-cpp/communication/RpcClient.h
@@ -62,10 +62,30 @@ struct RpcClient {
 	using MessageOrStatus =
 	    utils::Expected<v1::UMessage, std::variant<v1::UStatus, Commstatus>>;
 
+	/// @brief Callback function signature used in the callback form of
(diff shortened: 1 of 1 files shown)


gregmedd:
Forgive me: This change is very minor and will get scrutiny when #138 is implemented. I'm going to go forward with merging it.

PR #171, opened 2024-06-20
Add UT Coverage

MarkStokes71:
Added changes to makefile to create and capture UT coverage data.
Unit test themselves are excluded.
Also:
1. Implement workflow to measure test code coverage.
2. Generate coverage reports after tests are executed.
3. Add support for posting coverage comments on PRs from forks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ message("  Available build types are now: ${CMAKE_CONFIGURATION_TYPES}")
 message("* Current build type is: ${CMAKE_BUILD_TYPE}")
 
 # Settings for Gcc for all builds
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -Wall")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
 # Modify settings based on build type
 if(CMAKE_BUILD_TYPE STREQUAL Coverage)
@@ -40,7 +40,7 @@ message("* Current Compiler Flags: ${CMAKE_CXX_FLAGS}")
 #add_definitions(-DSPDLOG_FMT_EXTERNAL)
 
 # This is the root CMakeLists.txt file; We can set project wide settings here
-# TODO: Is this needed ? 
+# TODO: Is this needed? 
 if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
 	set(CMAKE_CXX_STANDARD 17)
 	# place libraries in a lib directory and executables in a bin directory,
@@ -64,7 +64,6 @@ target_include_directories(${PROJECT_NAME}
 	${spdlog_INCLUDE_DIR})
 
 set_property(TARGET ${PROJECT_NAME} PROPERTY POSITION_INDEPENDENT_CODE ON)
-
 
 target_link_libraries(${PROJECT_NAME} 
 	PRIVATE 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,30 @@ find_package(protobuf REQUIRED)
 find_package(spdlog REQUIRED)
 find_package(up-core-api REQUIRED)
 
+message("* Adding build types...")
+list(APPEND CMAKE_CONFIGURATION_TYPES Release Coverage)
+list(REMOVE_DUPLICATES CMAKE_CONFIGURATION_TYPES)
+set(CMAKE_CONFIGURATION_TYPES "${CMAKE_CONFIGURATION_TYPES}" CACHE STRING
+	"Add the code coverage configuration."
+	FORCE)
+message("  Available build types are now: ${CMAKE_CONFIGURATION_TYPES}")
+
+#Handle case where no build type is given (assume "Release")
+IF(NOT CMAKE_BUILD_TYPE)
+    SET(CMAKE_BUILD_TYPE Release
+        CACHE STRING "No Build type given, using default: Release"
+        FORCE)
+ENDIF(NOT CMAKE_BUILD_TYPE)
+message("* Current build type is: ${CMAKE_BUILD_TYPE}")
+
+# Settings for Gcc for all builds
+SET(GCC_BUILD_FLAGS "-g -Wall")
+
+# Modify settings based on build type
+if(CMAKE_BUILD_TYPE STREQUAL Coverage)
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -33,6 +34,7 @@ file(GLOB_RECURSE SRC_FILES "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp")
 
 add_library(${PROJECT_NAME} ${SRC_FILES})
 add_library(up-cpp::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}  -fprofile-arcs -ftest-coverage")
```
gregmedd, line 37: Should this be controlled with a build flag? We probably don't want to be producing coverage reports in production builds.

The gcovr documentation [recommends adding a "PROFILE" build type](https://www.gcovr.com/en/stable/cookbook.html#out-of-source-builds-with-cmake) to cmake for this (although "COVERAGE" might make more sense).

It looks like a `--coverage` flag was added to gcc as the new standard way to build and link for gcovr (this was news to me 😅 ). It's supposed in gcc 11.4 (our minimum supported version), so switching to that would be preferred.

MarkStokes71: This doesn't produce the ut report, this just prepares the output files needed to make such a report.
I tried the --coverage flag and it didn't work here.  I wanted to use it, but it produces an error, hence leaving what worked alone.

README.md
```diff
@@ -15,22 +15,26 @@ implementation, such as [up-transport-zenoh-cpp][zenoh-transport-repo].
 
 ### Requirements:
 - Compiler: GCC/G++ 11 or Clang 13
-- Conan : 1.59 or latest 2.X
+- pip : To install others
+- Conan(pip) : 1.59 or > 2.4.2
+- Cmake(apt) : > 3.22
+- Ninja(pip) : Latest
+- Gcovr(pip) : For UT code coverage if desired
```
gregmedd, line 22: Do not specify the package manager for installing tools. That will be different for every user, build environment, and distro. Developers are expected to know how to source and install tools for their environment.

README.md
```diff
@@ -15,22 +15,26 @@ implementation, such as [up-transport-zenoh-cpp][zenoh-transport-repo].
 
 ### Requirements:
 - Compiler: GCC/G++ 11 or Clang 13
-- Conan : 1.59 or latest 2.X
+- pip : To install others
+- Conan(pip) : 1.59 or > 2.4.2
```
gregmedd, line 19: Conan versions above 2.3.x are explicitly incompatible right now. Where is this Conan2 version number coming from?

I'm also not entirely sure conanfile.txt for this repo is still Conan1 compatible.

README.md
```diff
@@ -15,22 +15,26 @@ implementation, such as [up-transport-zenoh-cpp][zenoh-transport-repo].
...
+- Ninja(pip) : Latest
+- Gcovr(pip) : For UT code coverage if desired
 
 #### Conan packages
 
 Using the recipes found in [up-conan-recipes][conan-recipe-repo], build these
 Conan packages:
 
-1. [up-core-api][spec-repo] - `conan create --version 1.5.8 --build=missing up-core-api/developer`
+1. [up-core-api][spec-repo]: `conan create --version 1.5.8 --build=missing up-core-api/developer`
```
gregmedd, line 29: Relevance?

README.md
```diff
@@ -15,22 +15,26 @@ implementation, such as [up-transport-zenoh-cpp][zenoh-transport-repo].
...
 
 **NOTE:** all `conan` commands in this document use  Conan 2.x syntax. Please
 adjust accordingly when using Conan 1.x.
 
 ## How to Use the Library
 
 To add up-cpp to your conan build dependencies, place following in your
-conanfile.txt:
+conanfile.txt (https://docs.conan.io/2/reference/conanfile_txt.html):
```
gregmedd, line 37: Use markdown link syntax
```suggestion
[conanfile.txt][conanfile-txt-docs]:
```
and place the URL at the end of the file with the other URLs.

README.md
```diff
@@ -56,6 +60,8 @@ up-cpp, If you are making a project that uses up-cpp, follow the steps in the
 
 ### With Conan for dependencies
 
+Clone the up-cpp repo to a folder named: up-cpp
```
gregmedd, line 63: That's the default folder name when cloning a repo called up-cpp. Do we really need to document that?

gregmedd (changes requested):
I would recommend restricting the scope of changes to README.md to only the changes relevant to this PR. For additional cleanup, submit a separate bug+PR against the documentation.

README.md
MarkStokes71: Ya, I don't know.  This is what you mentioned in the Teams chat and what worked for me.  Also, I agree that we might remove the Conan 1.x comments.  However, I was chatting with someone on Yammer that couldn't get Conan 2.3.2 (His comment was: Normally we can only pull Python packages from Artifactory, and if it's not there, we have to get the open source team to scan and add it.).  I mentioned he could just get it from Github, but he said the CI's are going to need it so, he might as well get it figured out now anyway.  I'll go ahead and remove the 1.x stuff.

CMakeLists.txt
gregmedd: The `--coverage` flag is not the only topic discussed here. At a minimum a new build type needs to be added to prevent release builds from including all the cruft for generating coverage data.

README.md
gregmedd: Please don't change the requirements *at all* in the PR

README.md
MarkStokes71: The hyphen makes it look like it's part of a command.  The colon to me shows this is the command to create the Conan build.

README.md
MarkStokes71: The real question is, it is completely clear without it?  The very next instruction says cd to the up-cpp folder with no indication of how that folder was created.  Even if this is the default behavior, I always find it more helpful to be verbose.  Would you rather people unfamiliar w/ git and the default behavior be frustrated and submitting support tickets for what do these instructions mean?  Or just have the instructions be explicit?

README.md
MarkStokes71: Resolved.  Actually, three of those are platform independent (all installed w/ pip).

CMakeLists.txt
MarkStokes71: Done with "Coverage" build type

README.md
MarkStokes71: Done

gregmedd (approved):
LGTM - Thanks for getting this done

PR #176, opened 2024-06-22
Add RPC Server

agosh01:
This PR implements RPC server interface to up-cpp

### DIFF
--- a/include/up-cpp/communication/RpcServer.h
+++ b/include/up-cpp/communication/RpcServer.h
@@ -91,9 +91,22 @@ protected:
 	///            respond() is called. Note that the original request's TTL
 	///            may also still apply.
 	RpcServer(std::shared_ptr<transport::UTransport> transport,
-	          const v1::UUri& method,
 	          std::optional<v1::UPayloadFormat> format = {},
 	          std::optional<std::chrono::milliseconds> ttl = {});
+
+	/// @brief Allows std::make_unique to directly access RpcServer's private
+	/// constructor.
+	///
+	/// @param transport The transport layer abstraction for the RPC server.
+	/// @param payload_format (Optional) Specifies the payload format, if any.
+	/// @param ttl (Optional) Specifies the time-to-live (TTL) value, if any.
+	friend std::unique_ptr<RpcServer>
+	std::make_unique<RpcServer, std::shared_ptr<transport::UTransport>,
+	                 std::optional<v1::UPayloadFormat>,
+	                 std::optional<std::chrono::milliseconds>>(
+	    std::shared_ptr<uprotocol::transport::UTransport>&&,
+	    std::optional<uprotocol::v1::UPayloadFormat>&&,
+	    std::optional<std::chrono::milliseconds>&&);
 
 	/// @brief Connects the RPC callback method and returns the status from
 	///        UTransport::registerListener.
@@ -114,11 +127,11 @@ private:
 	/// @brief RPC callback method
 	RpcCallback callback_;
 
-	/// @brief Handle to the connected callback for the RPC method wrapper
-	transport::UTransport::ListenHandle callback_handle_;
-
 	/// @brief Format of the payload that will be expected in responses
 	std::optional<v1::UPayloadFormat> expected_payload_format_;
+
+	/// @brief Handle to the connected callback for the RPC method wrapper
+	transport::UTransport::ListenHandle callback_handle_;
 };
 
 }  // namespace uprotocol::communication
--- a/include/up-cpp/communication/RpcServer.h
+++ b/include/up-cpp/communication/RpcServer.h
@@ -15,6 +15,7 @@
 #include <up-cpp/datamodel/builder/Payload.h>
 #include <up-cpp/datamodel/builder/UMessage.h>
 #include <up-cpp/datamodel/validator/UMessage.h>
+#include <up-cpp/datamodel/validator/UUri.h>
 #include <up-cpp/transport/UTransport.h>
 #include <uprotocol/v1/umessage.pb.h>
 #include <uprotocol/v1/uri.pb.h>
@@ -100,7 +101,8 @@ protected:
 	/// @param callback Method that will be called when requests are received.
 	///
 	/// @returns OK if connected successfully, error status otherwise.
-	[[nodiscard]] v1::UStatus connect(RpcCallback&& callback);
+	[[nodiscard]] v1::UStatus connect(const v1::UUri& method,
+	                                  RpcCallback&& callback);
 
 private:
 	/// @brief Transport instance that will be used for communication
@@ -114,6 +116,9 @@ private:
 
 	/// @brief Handle to the connected callback for the RPC method wrapper
 	transport::UTransport::ListenHandle callback_handle_;
+
+	/// @brief Format of the payload that will be expected in responses
+	std::optional<v1::UPayloadFormat> expected_payload_format_;
 };
 
 }  // namespace uprotocol::communication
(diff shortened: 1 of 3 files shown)


include/up-cpp/communication/RpcServer.h
```diff
@@ -114,6 +116,9 @@ struct RpcServer {
 
 	/// @brief Handle to the connected callback for the RPC method wrapper
 	transport::UTransport::ListenHandle callback_handle_;
+
+	/// @brief Format of the payload that will be expected in responses
+	std::optional<v1::UPayloadFormat> expected_payload_format_;
```
gregmedd, line 121: Minor: move this before the `callback_handle_` member. This ensures the handle is destroyed before any other members are deallocated.

src/communication/RpcServer.cpp
```diff
@@ -10,3 +10,125 @@
...
+		v1::UStatus status;
+		status.set_code(v1::UCode::OUT_OF_RANGE);
+		status.set_message(e.what());
+		return uprotocol::utils::Unexpected(status);
+	} catch (const std::exception& e) {
+		v1::UStatus status;
+		status.set_code(v1::UCode::UNKNOWN);
+		status.set_message(e.what());
+		return uprotocol::utils::Unexpected(status);
+	}
```
gregmedd, line 63: We don't use try/catch for flow control. If an exception was thrown, it is a panic that should be seen by the developer.

Instead, the only UStatus that should be returned by this function should be the one returned by `transport->registerListener()` (via `RpcServer::connect()`).

Also, don't use `new` directly. Instead, favor `std::make_unique`.

```suggestion
	auto server = std::make_unique<RpcServer>(transport, method_name, payload_format, ttl);
```

src/communication/RpcServer.cpp
```diff
@@ -10,3 +10,125 @@
...
+	} else {
+		return uprotocol::utils::Unexpected(std::move(status));
+	}
+}
+
+v1::UStatus RpcServer::connect(const v1::UUri& method, RpcCallback&& callback) {
+	callback_ = std::move(callback);
+	auto result = transport_->registerListener(
+	    /*sink_filter=*/{method},
+	    /*listener=*/[this](const v1::UMessage& message) {
```
gregmedd, line 77: Minor: prefer C++ style comments where possible:
```suggestion
	    // sink_filter=
	    method,
	    // listener=
	    [this](const v1::UMessage& message) {
```

gregmedd: I would also recommend renaming `message` either `request` or `request_message` in this lambda for readability

src/communication/RpcServer.cpp
```diff
@@ -10,3 +10,125 @@
...
+		    if (!valid) {
+			    auto response =
+			        datamodel::builder::UMessageBuilder::response(message)
+			            .withCommStatus(v1::UCode::INVALID_ARGUMENT)
+			            .build({std::string(Validator::message::message(
+			                        reason.value())),
+			                    v1::UPayloadFormat::UPAYLOAD_FORMAT_TEXT});
+
+			    // Ignoring status code for transport send
+			    auto _ = transport_->send(response);
```
gregmedd, line 90: I just saw Mark use [`std::ignore`](https://en.cppreference.com/w/cpp/utility/tuple/ignore) for this purpose in another PR. It's new to me, but might read better:
```suggestion
			    std::ignore = transport_->send(response);
```

src/communication/RpcServer.cpp
```diff
@@ -10,3 +10,125 @@
...
+			    return;
+		    }
+
+		    auto payloadData = callback_(message);
+		    auto builder =
+		        datamodel::builder::UMessageBuilder::response(message);
```
gregmedd, line 96: Minor: if this were moved to around 78, the same builder could be used for either the invalid request or valid request paths. It would probably help with cleaning up the code inside the `if (!valid)` block

src/communication/RpcServer.cpp
```diff
@@ -10,3 +10,125 @@
...
+		    // Extract payload data and format from the callback
+		    auto [valid, reason] =
+		        Validator::message::isValidRpcRequest(message);
+		    if (!valid) {
+			    auto response =
+			        datamodel::builder::UMessageBuilder::response(message)
+			            .withCommStatus(v1::UCode::INVALID_ARGUMENT)
+			            .build({std::string(Validator::message::message(
+			                        reason.value())),
+			                    v1::UPayloadFormat::UPAYLOAD_FORMAT_TEXT});
```
gregmedd, line 87: Minor: If the builder is declared before this block somewhere, this could be cleaned up like so:
```suggestion
			    auto message = static_cast<std::string>(Validator::message::message(reason.value()));
			    auto response = builder.withCommStatus(v1::UCode::INVALID_ARGUMENT)
			            .build({std::move(message), v1::UPayloadFormat::UPAYLOAD_FORMAT_TEXT});
```

test/coverage/communication/RpcServerTest.cpp
```diff
@@ -9,32 +9,400 @@
...
+#include <sstream>
+
 #include "UTransportMock.h"
 
+namespace uprotocol::test {
+uprotocol::test::UTransportMockInfo UTransportMock::mock_info =
+    uprotocol::test::UTransportMockInfo();
+};
+
+auto& mock_info = uprotocol::test::UTransportMock::mock_info;
```
gregmedd, line 27: This will be removed when you rebase to main

gregmedd (changes requested):
I've listed a few changes that need to be made. I'm still reading through the test code

src/communication/RpcServer.cpp
agosh01: yes originally U tried to use std::make_unique compiler throws error that it can't see constructor for this class as its private therefore using manual way which should be same 

 Compiler error:
 
```
 'uprotocol::communication::RpcServer::RpcServer(std::shared_ptr<uprotocol::transport::UTransport>, std::optional<uprotocol::v1::UPayloadFormat>, std::optional<std::chrono::duration<long int, std::ratio<1, 1000> > >)' is protected within this context
 1070 |     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }

```

include/up-cpp/communication/RpcServer.h
agosh01: addressed

src/communication/RpcServer.cpp
agosh01: addressed

src/communication/RpcServer.cpp
agosh01: addressed

src/communication/RpcServer.cpp
agosh01: addressed, thanks for suggestion

src/communication/RpcServer.cpp
agosh01: addressed thanks for suggestion

test/coverage/communication/RpcServerTest.cpp
agosh01: addressed

src/communication/RpcServer.cpp
```diff
@@ -10,3 +10,116 @@
...
+		    // presence
+		    if (!payloadData.has_value()) {
+			    // builder.build() verifies if payload format is required
+			    auto response = builder.build();
+			    // Ignoring status code for transport send
+			    std::ignore = transport_->send(response);
+		    } else {
+			    // builder.build(payloadData) verifies if payload format
+			    // matches the expected
+			    auto response = builder.build(std::move(payloadData.value()));
```
gregmedd, line 107: Minor: I _think_ this needs to be
```suggestion
			    auto response = builder.build(std::move(payloadData).value());
```
in order for the move to actually work.

gregmedd (approved):
I only see one minor change and then this is good to go

src/communication/RpcServer.cpp
agosh01: Done

PR #177, opened 2024-06-24
Refactor UTransportMock to remove lifecycle testing, and so make suitable for inclusion in other tests.

debruce:
init/deinit tesing has been removed, and so the UTransportMock no longer needs to contain a static member for checking for deinitialization. Furthermore, the memory leak detector was removed, since this is supposed to be cover elsewhere by valgrind.


gregmedd:
Could the mock info object just be a regular member in UTransportMock? Does it need to be a singleton or static at all?

PR #180, opened 2024-06-25
Update UMessage validator's isValid() to return better reasons

### LINKED ISSUE #178: UMessage validator isValid() will always return WRONG_MESSAGE_TYPE for invalid (non-notification) messages
The current implementation of `isValid()` checks validity for each message type, returning true on the first valid message. At the end, it returns whatever is returned by `isValidNotification()`. For any message that is not of notification type, the reason returned will always be `WRONG_MESSAGE_TYPE`. This is not particularly helpful in diagnosing an error with a message.

Since a message can only have one type, we can assume that any reason that is not `WRONG_MESSAGE_TYPE` is probably the _right_ reason the message failed validation.

gregmedd:
`isValid()` would often only return the `WRONG_MESSAGE_TYPE` reason when there was a better reason it could return. Adds test to catch this issue and fixes the returned reasons.

closes #178

### DIFF
--- a/src/datamodel/validator/UMessage.cpp
+++ b/src/datamodel/validator/UMessage.cpp
@@ -31,9 +31,6 @@ std::string_view message(Reason reason) {
 			return "The Priority, if set, is not within the allowable range";
 		case Reason::PAYLOAD_FORMAT_OUT_OF_RANGE:
 			return "The Payload Format is not within the allowable range";
-		case Reason::WRONG_MESSAGE_TYPE:
-			return "The type set in the message is incorrect for the validated "
-			       "mode";
 		case Reason::BAD_SOURCE_URI:
 			return "Source URI did not pass validity checks";
 		case Reason::BAD_SINK_URI:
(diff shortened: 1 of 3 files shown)


test/coverage/datamodel/UMessageValidatorTest.cpp
```diff
@@ -770,4 +770,94 @@ TEST_F(TestUMessageValidator, ValidNotification) {
 	}
 }
 
+TEST_F(TestUMessageValidator, IsValid) {
```
billpittman, line 773: Good Catch - totally missed testing isValid.

test/coverage/datamodel/UMessageValidatorTest.cpp
```diff
@@ -770,4 +770,94 @@ TEST_F(TestUMessageValidator, ValidNotification) {
...
+
+	{
+		// valid response
+		auto source = source_;
+		auto sink = sink_;
+		source.set_resource_id(0);
+
+		auto attributes = fakeResponse(source, sink);
+		auto umessage = build(attributes);
+		auto [valid, reason] = isValidRpcResponse(umessage);
```
billpittman, line 804: Function being called: isValidRpcResponse -> isValid

test/coverage/datamodel/UMessageValidatorTest.cpp
```diff
@@ -770,4 +770,94 @@ TEST_F(TestUMessageValidator, ValidNotification) {
...
+	}
+
+	{
+		// valid publish
+		auto source = source_;
+		source.set_resource_id(0x8000);
+
+		auto attributes = fakePublish(source);
+		auto umessage = build(attributes);
+		auto [valid, reason] = isValidPublish(umessage);
```
billpittman, line 825: Function being called: IsValidPublish -> isValid

Honestly both wouldn't hurt, but that is a bit of over-testing in my opinion.

billpittman (changes requested):

src/datamodel/validator/UMessage.cpp
```diff
@@ -50,34 +47,37 @@ std::string_view message(Reason reason) {
...
 			       "not been swapped";
+		case Reason::WRONG_MESSAGE_TYPE:
+			return "The type set in the message is incorrect for the validated "
+			       "mode";
 		default:
 			return "Unknown reason.";
 	}
 }
 
 ValidationResult isValid(const v1::UMessage& umessage) {
```
billpittman, line 58: I wonder if we'd have a better "isValid" with better error codes if we respect the umessage.attributes.type() field ... like this:

`        if (umessage.attributes().type() ==
            v1::UMessageType::UMESSAGE_TYPE_REQUEST) {
                return isValidRpcRequest(message);
        }
`

That way if a message says "I'm a request" you only test for that?  Since every other type will fail with WRONG_MESSAGE_TYPE.

test/coverage/datamodel/UMessageValidatorTest.cpp
gregmedd: Good catch. Fixed

test/coverage/datamodel/UMessageValidatorTest.cpp
gregmedd: Good catch. Fixed

src/datamodel/validator/UMessage.cpp
gregmedd: That's not a bad idea :thinking: 

It would also be more efficient. I'll give it a try

gregmedd: Done

billpittman (approved):
isValid looks a lot better now.  I think this should resolve the issue or get us much much closer at the very least.

PR #185, opened 2024-06-26
UUID: Convert UUID8 to UUID7

billpittman:
Switch to UUID version 7 and refactor tests.


src/datamodel/serializer/Uuid.cpp
```diff
@@ -17,22 +17,10 @@
...
-constexpr uint64_t VARIANT_MASK = 0x3;
-constexpr uint64_t VARIANT_SHIFT = 62;
-constexpr uint64_t RANDOM_MASK = 0x3FFFFFFFFFFFFFFF;
 constexpr uint64_t RANDOM_SHIFT = 48;
-constexpr size_t UUID_BYTE_SIZE = 16;
-constexpr size_t UUID_PART_SIZE = 4;
-constexpr uint32_t HEX_BASE = 16;
-constexpr uint64_t MASK_32_BITS = 0xFFFFFFFF;
-constexpr uint64_t MASK_16_BITS = 0xFFFF;
-constexpr uint64_t MASK_14_BITS = 0x3FFF;
```
gregmedd, line 35: :heart: nice to see these swapped to using the constants header.

src/datamodel/serializer/Uuid.cpp
```diff
@@ -42,35 +30,38 @@ constexpr size_t LSB_LOW__ = 12;
...
-	uint64_t unix_ts_ms = (uuid.msb() >> TIMESTAMP_SHIFT) & TIMESTAMP_MASK;
-	uint8_t ver = (uuid.msb() >> VERSION_SHIFT) & VERSION_MASK;
-	uint16_t counter = uuid.msb() & COUNTER_MASK;
-	uint8_t var = (uuid.lsb() >> VARIANT_SHIFT) & VARIANT_MASK;
-	uint64_t rand_b = uuid.lsb() & RANDOM_MASK;
+	// Extracting the parts of the UUIDv7
+	uint64_t unix_ts_ms =
+	    (uuid.msb() >> UUID_TIMESTAMP_SHIFT) & UUID_TIMESTAMP_MASK;
+	uint8_t ver = (uuid.msb() >> UUID_VERSION_SHIFT) & UUID_VERSION_MASK;
+	uint16_t counter = uuid.msb() & UUID_COUNTER_MASK;
```
gregmedd, line 37: Minor: Since UUIDv8 calls this field `rand_a`, can you change the name of the variables related to that field? I think it will help with reading this code against the UUID RFC in the future.

include/up-cpp/datamodel/constants/UuidConstants.h
```diff
@@ -25,8 +25,8 @@ constexpr uint64_t UUID_COUNTER_MASK = 0xFFF;
 constexpr uint64_t UUID_RANDOM_MASK = 0x3FFFFFFFFFFFFFFF;
 
 // Constants for UUID version and variant
-constexpr uint8_t UUID_VERSION_8 = 8;
-constexpr uint8_t UUID_VARIANT_RFC4122 = 2;
+constexpr uint64_t UUID_VERSION_7 = 7;
+constexpr uint64_t UUID_VARIANT_RFC4122 = 2;
```
gregmedd, line 29: Is this variant correct? At least in [RFC9562](https://www.rfc-editor.org/rfc/rfc9562#section-4.1), it looks like it should actually be in the range `[0x8, 0xB]`

src/datamodel/validator/Uuid.cpp
```diff
@@ -65,12 +60,12 @@ ValidationResult isUuid(const uprotocol::v1::UUID uuid) {
...
 
 	uint8_t version = internalGetVersion(uuid);
-	if (version != 8) {
+	if (version != UUID_VERSION_7) {
 		return {false, Reason::WRONG_VERSION};
 	}
 
 	uint8_t variant = internalGetVariant(uuid);
-	if (variant != 2) {  // Variant should be 10 in binary
+	if (variant != UUID_VARIANT_RFC4122) {  // Variant should be 10 in binary
```
gregmedd, line 68: Minor:
```suggestion
	if (variant != UUID_VARIANT_RFC4122) {
```

test/coverage/datamodel/UMessageBuilderTest.cpp
```diff
@@ -76,8 +77,7 @@ class TestUMessageBuilder : public testing::Test {
 		sink_.set_ue_version_major(0xF9);
 		sink_.set_resource_id(0x8102);
 
-		reqId_.set_msb((8ULL << 12) | (0x123ULL));
-		reqId_.set_lsb((2ULL << 62) | (0xFFFFFFFFFFFFULL));
+		reqId_ = UuidBuilder::getBuilder().build();
```
gregmedd, line 80: :heart: It's like you read my mind

test/coverage/datamodel/UuidBuilderTest.cpp
gregmedd, line 1: Based on the comment in `builder/Uuid.cpp`, I think it's likely there will be some changes to the builder tests. Mostly, I expect a few tests to be removed, particularly if the shared state isn't needed anymore. That would also fix #187

gregmedd (changes requested):
Overall, the direction of this change looks good to me. There are some builder changes required to address the differences between UUIDv8 and UUIDv7 (mostly that the random fields are generated uniquely for every `build()` call).

gregmedd:
This PR will also need to be rebased to main to resolve merge conflicts.

src/datamodel/builder/Uuid.cpp
```diff
@@ -52,50 +48,23 @@ UuidBuilder& UuidBuilder::withRandomSource(
...
 	    std::chrono::time_point_cast<std::chrono::milliseconds>(now);
-
-	if (unix_ts_ms != shared_state_->last_unix_ts_ms) {
-		// Reset the counter if the timestamp tick has changed
-		shared_state_->counter = 0;
-		shared_state_->last_unix_ts_ms = unix_ts_ms;
-	}
+	std::mt19937_64 gen{std::random_device{}()};
+	std::uniform_int_distribution<uint16_t> dist_a(0, UUID_RANDOM_A_MASK);
+	std::uniform_int_distribution<uint64_t> dist_b(0, UUID_RANDOM_B_MASK);
```
gregmedd, line 58: Minor: make these variables members of the `UuidBuilder` class. They can then be reused for each call to build and, if the `UuidBuilder` instance is held, the caller will get a little bit of performance benefit on each call to `build()`.

gregmedd (approved):
TBH the only two comments I have remaining are pretty minor. Let's take those as follow-up tasks for after the alpha.2 release.

PR #170, opened 2024-06-20
Publisher Implementation

ruchirchauhan:
Implements communication/publisher functionality

PR #186, opened 2024-06-26
Notificationsource Implementation

ruchirchauhan:
Implements NotificationSource


gregmedd:
Giving this a quick look over, it appears to be very similar to Publisher (#170). So long as comments from there are applied here too, I do not see a reason this could not be approved and merged at the same time.

gregmedd (approved):
LGTM

Issue #188, opened 2024-06-26
Confirm the UMessage URI checks align with spec


gregmedd:
> Looking at the [UAttributes spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc#2-common-message-attributes), I think we should be checking both the source and sink against specific rules for each message type. However, we may also need additional UUri validators for this. I think I'll file a separate bug for this so it can be investigated more thoroughly.

_Originally posted by @gregmedd in https://github.com/eclipse-uprotocol/up-cpp/pull/184#discussion_r1655347813_

----

The spec indicates that in some message types, there are specific rules for both source _and_ sink URIs. Currently only one of those is checked against specific rules in the UMessage validator code.

billpittman:
I will take a look at this issue.

PR #200, opened 2024-07-02
UMessage: Add more notification UUri checking

billpittman:
Split the uuri::isValidNotification function into two functions, one for the source and one for sink and update appropriate tests/files.

This PR is intended to close issue #188

### DIFF
--- a/src/datamodel/validator/UMessage.cpp
+++ b/src/datamodel/validator/UMessage.cpp
@@ -327,7 +327,7 @@ ValidationResult isValidNotification(const v1::UMessage& umessage) {
 
 	{
 		auto [valid, reason] =
-		    uri::isValidNotification(umessage.attributes().source());
+		    uri::isValidNotificationSource(umessage.attributes().source());
 		if (!valid) {
 			return {false, Reason::BAD_SOURCE_URI};
 		}
@@ -335,7 +335,7 @@ ValidationResult isValidNotification(const v1::UMessage& umessage) {
 
 	{
 		auto [valid, reason] =
-		    uri::isValidNotification(umessage.attributes().sink());
+		    uri::isValidNotificationSink(umessage.attributes().sink());
 		if (!valid) {
 			return {false, Reason::BAD_SINK_URI};
 		}
--- a/src/datamodel/validator/UUri.cpp
+++ b/src/datamodel/validator/UUri.cpp
@@ -75,7 +75,14 @@ ValidationResult isValid(const v1::UUri& uuri) {
 		}
 	}
 
-	return isValidNotification(uuri);
+	{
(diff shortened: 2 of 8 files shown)


src/datamodel/validator/UMessage.cpp
```diff
@@ -339,6 +342,9 @@ ValidationResult isValidNotification(const v1::UMessage& umessage) {
 		if (!valid) {
 			return {false, Reason::BAD_SINK_URI};
 		}
+		if (umessage.attributes().sink().resource_id() != 0) {
+			return {false, Reason::BAD_SINK_URI};
+		}
```
gregmedd, line 347: These checks should be in the URI validator. I would recommend splitting `isValidNotificaton()` into `isValidNotificationSource()` and `isValidNotificationSink()`. RPC already has separate source/sink validators (called request and response), and publish only uses one URI, so I think notifications are the only one that need the split.

src/datamodel/validator/UUri.cpp
```diff
@@ -109,7 +109,15 @@ ValidationResult isValidDefaultSource(const v1::UUri& uuri) {
...
 	}
 
-	return isValidRpcResponse(uuri);
+	if (uses_wildcards(uuri)) {
+		return {false, Reason::DISALLOWED_WILDCARD};
+	}
+
+	if (uuri.resource_id() == 0) {
+		return {false, Reason::BAD_RESOURCE_ID};
+	}
```
gregmedd, line 118: Using `0` as the default source's resource ID was intentional - it is unset and needs to be set when used (i.e. should fail validation when used in a communication module).

I've actually been considering adding an interface to the L2 communication classes that _only_ takes resource ID instead of the whole source URI to make this clearer to the user of the interface. Something like:

```
	NotificationSource(std::shared_ptr<transport::UTransport> transport,
	                   uint16_t source_resource_id, v1::UUri&& sink,
	                   std::optional<v1::UPayloadFormat> payload_format = {},
	                   std::optional<v1::UPriority> priority = {},
	                   std::optional<std::chrono::milliseconds> ttl = {}) {

		auto source_uri = transport->getDefaultSource();
		source_uri.set_resource_id(source_resource_id);
		...
	}
```

We would still need to make sure the URIs are being validated correctly in the message builder, but I think this would make the expectation more clear at the communication layer.

billpittman: So if it was intentionally set to zero, should the UUri validator validate that its zero and nothing else?  IE Change the check to != 0 ?  Or are we not wanting to be so deliberate?

MarkStokes71 (approved):

gregmedd (approved):
LGTM - my only questions need to be directed at the spec itself.

gregmedd:
Commit history looks good - merging

PR #202, opened 2024-07-09
Implementation of RpcClient

gregmedd:
`RpcClient` implements the client side of the layer 2 RPC model. It must be able to:

1. Send well-formed request messages to a listening RPC server
2. Map responses back to the original caller using the request's message UUID
3. Provide response messages to the caller
4. Expire pending requests that have been waiting longer than the request TTL

----

When invoking an RPC method, the sequence of events is as follows:

- A request message is generated
- A listener wrapper for the caller's callback (or promise) is generated. This listener will check the request ID in the response against the ID from the original request, only passing the response onward when they match
- An expiration wrapper for the caller's callback is generated
- The listener wrapper is registered with the transport for the entity's RPC response URI. If an error is reported, the expiration wrapper is called and the error status is passed to the caller
- If the listener is registered successfully, the request is sent. If an error is reported, the expiration wrapper is called and the error status is passed to the caller
- If the request is sent successfully, then the expiration wrapper is queued with the thread that monitors for timeout events

A single shared thread (`ExpireWorker`) monitors a priority queue of pending requests. The queue is sorted such that the earliest expiration is always at the top of the queue. When entries are added to the queue, a condition variable wakes the worker. The worker sleeps until the next expiration time. If the expiring request has not already been fulfilled, then the worker will call the expiration wrapper to inform the waiting caller that the RPC request has timed out.

### DIFF
--- a/include/up-cpp/communication/RpcClient.h
+++ b/include/up-cpp/communication/RpcClient.h
@@ -67,9 +67,44 @@ struct RpcClient {
 	///        on the commstatus attributes field in any returned messages.
 	using MessageOrStatus = utils::Expected<v1::UMessage, Status>;
 
+	/// @brief Callback connections for callbacks passed to invokeMethod()
+	using Connection = utils::callbacks::Connection<void, MessageOrStatus&&>;
+
 	/// @brief Callback function signature used in the callback form of
-	///        invokeMethod
-	using Callback = std::function<void(MessageOrStatus)>;
+	///        invokeMethod()
+	using Callback = Connection::Callback;
+
+	/// @brief Handle that must be held for a callback to remain registered
+	///        for the duration of an RPC call.
+	using InvokeHandle = Connection::Handle;
+
+	/// @brief Extension to std::future that also holds a callback handle
+	class InvokeFuture {
+		InvokeHandle callback_handle_;
+		std::future<MessageOrStatus> future_;
+
+	public:
+		InvokeFuture();
+		InvokeFuture(InvokeFuture&&) noexcept;
+		InvokeFuture(std::future<MessageOrStatus>&&, InvokeHandle&&) noexcept;
+
+		InvokeFuture& operator=(InvokeFuture&&) noexcept;
+
+		/// @name Passthroughs for std::future
+		/// @{
+		auto get() { return future_.get(); }
+		auto valid() const noexcept { return future_.valid(); }
+		void wait() const { future_.wait(); }
+		template <typename... Args>
+		auto wait_for(Args&&... args) const {
+			return future_.wait_for(std::forward<Args>(args)...);
+		}
+		template <typename... Args>
+		auto wait_until(Args&&... args) const {
+			return future_.wait_until(std::forward<Args>(args)...);
+		}
+		/// @}
+	};
 
 	/// @brief Invokes an RPC method by sending a request message.
 	///
@@ -83,7 +118,8 @@ struct RpcClient {
 	///       * A UStatus with the value returned by UTransport::send().
 	///       * A Commstatus as received in the response message (if not OK).
 	///       * A UMessage containing the response from the RPC target.
-	void invokeMethod(datamodel::builder::Payload&&, Callback&&);
+	[[nodiscard]] InvokeHandle invokeMethod(datamodel::builder::Payload&&,
+	                                        Callback&&);
 
 	/// @brief Invokes an RPC method by sending a request message.
 	///
@@ -98,8 +134,7 @@ struct RpcClient {
 	///          * A UStatus with the value returned by UTransport::send().
 	///          * A Commstatus as received in the response message (if not OK).
 	///          * A UMessage containing the response from the RPC target.
-	[[nodiscard]] std::future<MessageOrStatus> invokeMethod(
-	    datamodel::builder::Payload&&);
+	[[nodiscard]] InvokeFuture invokeMethod(datamodel::builder::Payload&&);
 
 	/// @brief Invokes an RPC method by sending a request message.
 	///
@@ -114,7 +149,7 @@ struct RpcClient {
 	///       * A UStatus with the value returned by UTransport::send().
 	///       * A Commstatus as received in the response message (if not OK).
 	///       * A UMessage containing the response from the RPC target.
-	void invokeMethod(Callback&&);
+	[[nodiscard]] InvokeHandle invokeMethod(Callback&&);
 
 	/// @brief Invokes an RPC method by sending a request message.
 	///
@@ -129,7 +164,7 @@ struct RpcClient {
 	///          * A UStatus with the value returned by UTransport::send().
 	///          * A Commstatus as received in the response message (if not OK).
 	///          * A UMessage containing the response from the RPC target.
-	[[nodiscard]] std::future<MessageOrStatus> invokeMethod();
+	[[nodiscard]] InvokeFuture invokeMethod();
 
 	/// @brief Default move constructor (defined in RpcClient.cpp)
 	RpcClient(RpcClient&&);
@@ -140,7 +175,7 @@ struct RpcClient {
 private:
 	/// @brief Internal implementation of invokeMethod that handles all the
 	///        shared logic for the public invokeMethod() methods.
-	void invokeMethod(v1::UMessage&&, Callback&&);
+	InvokeHandle invokeMethod(v1::UMessage&&, Callback&&);
 
 	/// @brief Handle to a shared worker that monitors for and cancels expired
 	///        requests.
--- a/include/up-cpp/communication/RpcClient.h
+++ b/include/up-cpp/communication/RpcClient.h
@@ -54,13 +54,18 @@ struct RpcClient {
 	                   std::optional<uint32_t> permission_level = {},
 	                   std::optional<std::string> token = {});
 
+	/// @brief A status code that can be returned in the response message's
+	///        attributes.
 	using Commstatus = v1::UCode;
 
+	/// @brief Represents the combined status possibilities for transport and
+	///        uProtocol network communications.
+	using Status = std::variant<v1::UStatus, Commstatus>;
+
 	/// @brief Contains either a UMessage (when successful) or a UStatus /
-	/// Commstatus when an error occurred. Commstatus would be set based
-	/// on the commstatus attributes field in any returned messages.
-	using MessageOrStatus =
-	    utils::Expected<v1::UMessage, std::variant<v1::UStatus, Commstatus>>;
+	///        Commstatus when an error occurred. Commstatus would be set based
+	///        on the commstatus attributes field in any returned messages.
+	using MessageOrStatus = utils::Expected<v1::UMessage, Status>;
 
 	/// @brief Callback function signature used in the callback form of
 	///        invokeMethod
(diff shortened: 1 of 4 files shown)


gregmedd:
A note on coverage: the unit tests cover all but one line in `RpcClient.cpp`. This line is in the destructor for the class that wraps the priority queue where it cancels any pending requests. Pending requests for each individual `RpcClient` instance are cleaned up when the instance is removed, so by the time the priority queue is destroyed, none should remain. It might be possible to exercise this code by intentionally leaking an `RpcClient` instance in such a way that its destructor does not run, but I'm not sure that it's critical to test that line for the pending alpha.2 release.

stevenhartley (commented):
Please see comments

include/up-cpp/communication/RpcClient.h
```diff
@@ -54,17 +54,57 @@ struct RpcClient {
...
+	using Status = std::variant<v1::UStatus, Commstatus>;
+
 	/// @brief Contains either a UMessage (when successful) or a UStatus /
-	/// Commstatus when an error occurred. Commstatus would be set based
-	/// on the commstatus attributes field in any returned messages.
-	using MessageOrStatus =
-	    utils::Expected<v1::UMessage, std::variant<v1::UStatus, Commstatus>>;
+	///        Commstatus when an error occurred. Commstatus would be set based
+	///        on the commstatus attributes field in any returned messages.
+	using MessageOrStatus = utils::Expected<v1::UMessage, Status>;
```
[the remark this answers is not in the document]
gregmedd: UMessage since the payload type information is stored in the attributes. The payload on its own is not self-describing.

include/up-cpp/communication/RpcClient.h
```diff
@@ -54,17 +54,57 @@ struct RpcClient {
...
+
+	/// @brief Callback connections for callbacks passed to invokeMethod()
+	using Connection = utils::callbacks::Connection<void, MessageOrStatus&&>;
```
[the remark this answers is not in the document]
gregmedd: Yes. We have been asked by several C++ developers to provide a fully asynchronous interface that does not require them to implement a separate waiting thread for the futures. Since the native format for handling incoming messages on the transport is a callback, that option is provided in this interface.

It's worth noting that C++ futures do not work the same way as futures in Java. There are scenarios where they may not be acceptable.

include/up-cpp/communication/RpcClient.h
```diff
@@ -54,17 +54,57 @@ struct RpcClient {
 	                   std::optional<uint32_t> permission_level = {},
 	                   std::optional<std::string> token = {});
 
+	/// @brief A status code that can be returned in the response message's
+	///        attributes.
 	using Commstatus = v1::UCode;
 
+	/// @brief Represents the combined status possibilities for transport and
+	///        uProtocol network communications.
```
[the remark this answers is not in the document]
gregmedd: We could return only a UStatus by packing the commstatus into a UStatus if that would simplify things. When a message is returned with commstatus set, would the payload of the message be expected to contain a string describing the error (that is, does it contain the message for the UStatus)?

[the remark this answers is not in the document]
gregmedd: > We simply need to complete the future with an exception

Two notes on this:

1. We allow for either a callback or a future depending on the calling implementation's needs
2. We use `Expected` to return either success or failure to callers. _exceptions_ in our error model for C++ are a _panic condition_. C++ futures will throw an exception if they are not completed successfully, violating our error model.

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+		builder_.withToken(*token);
+	}
+}
+
+RpcClient::InvokeHandle RpcClient::invokeMethod(v1::UMessage&& request,
+                                                Callback&& callback) {
+	auto when_expire = std::chrono::steady_clock::now() + ttl_;
+	auto reqid = request.attributes().id();
+	// Used to ensure that, no matter what, the callback is only called once
+	auto callback_once = std::make_shared<std::once_flag>();
```
billpittman, line 116: Please improve this comment to include the fact that there is a race condition between the "expire logic" and the "received response" logic.  As it stands it almost implies that you don't know (or aren't sure) why some weird race condition exists.

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+
+auto PendingRequest::operator>(const PendingRequest& other) const {
+	return when_expire > other.when_expire;
+}
+
+ScrubablePendingQueue::~ScrubablePendingQueue() {
+	const v1::UStatus cancel_reason = []() {
+		v1::UStatus reason;
+		reason.set_code(v1::UCode::CANCELLED);
+		reason.set_message("ExpireWorker shutting down");
```
gregmedd, line 247: Make this something more alarming since this list should never have anything in it - could label it as a possible memory leak

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+	                   }),
+	    c.end());
+
+	// TODO - is there a better way to shrink the internal container?
+	// Maybe instead we should enforce a capacity limit
+	constexpr size_t capacity_shrink_threshold = 16;
+	if ((c.capacity() > capacity_shrink_threshold) &&
+	    (c.size() < c.capacity() / 2)) {
+		c.shrink_to_fit();
+	}
```
gregmedd, line 279: Maybe `std::deque`? File an issue to follow up

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+		const auto now = std::chrono::steady_clock::now();
+		std::optional<decltype(PendingRequest::expire)> maybe_expire;
+
+		{
+			transport::UTransport::ListenHandle expired_handle;
+			std::lock_guard lock(pending_mtx_);
+			if (!pending_.empty()) {
+				const auto when_expire = pending_.top().when_expire;
+				if (when_expire <= now) {
+					maybe_expire = pending_.top().expire;
```
gregmedd, line 338: could std::move() this

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+				auto wake_when = pending_.top().when_expire;
+				wake_worker_.wait_until(lock, wake_when, [this, &wake_when]() {
+					auto when_next_wake = wake_when;
+					if (!pending_.empty()) {
+						when_next_wake =
+						    std::min(wake_when, pending_.top().when_expire);
+					}
+					return stop_ ||
+					       (std::chrono::steady_clock::now() >= when_next_wake);
+				});
```
gregmedd, line 373: Comment to explain this better

gregmedd: Does this even make sense? It doesn't wake if the queue is empty but it could. And why is it taking the min of wake_when or pending_.top()?

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+			if (pending_.empty()) {
+				wake_worker_.wait(
+				    lock, [this]() { return stop_ || !pending_.empty(); });
+			} else {
+				auto wake_when = pending_.top().when_expire;
+				wake_worker_.wait_until(lock, wake_when, [this, &wake_when]() {
+					auto when_next_wake = wake_when;
+					if (!pending_.empty()) {
+						when_next_wake =
+						    std::min(wake_when, pending_.top().when_expire);
```
gregmedd, line 369: This is a bug. It should return true if wake when and and pending top have different times. Otherwise this will go back to sleep for the original timeout even if the next expiring request is earlier.

test/coverage/communication/RpcClientTest.cpp
gregmedd, line 1: Schedule a PR to re-review this test

billpittman (approved):

gregmedd (commented):
Fix the bug found in review and send out for update.

src/communication/RpcClient.cpp
gregmedd: Thanks @billpittman for the catch!

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+			expire(expire_reason);
+
+		} else {
+			std::unique_lock lock(pending_mtx_);
+			if (pending_.empty()) {
+				wake_worker_.wait(
+				    lock, [this]() { return stop_ || !pending_.empty(); });
+			} else {
+				auto wake_when = pending_.top().when_expire;
+				wake_worker_.wait_until(lock, wake_when, [this, &wake_when]() {
```
billpittman, line 365: The pending_ list can have elements who's responses have been met, yet stay in this list until they "expire" and then attempt (but fail due to the call_once flag) to call the callback.  The issue is that they are in the list the entire time of their TTL even if the RPC request completes "instantly".

Right before we "wait_until", we _could_ decide to clean those items out of the list if the wait_until time is far enough in the future.

src/communication/RpcClient.cpp
```diff
@@ -10,3 +10,371 @@
...
+};
+
+}  // namespace detail
+}  // namespace
+
+namespace uprotocol::communication {
+
+////////////////////////////////////////////////////////////////////////////////
+struct RpcClient::ExpireService {
+	ExpireService() : instance_id_(next_instance_id++) {}
```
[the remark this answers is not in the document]
gregmedd: Issue #218 has been filed to follow-up on this for alpha.3

src/communication/RpcClient.cpp
gregmedd: Done

src/communication/RpcClient.cpp
gregmedd: Logged in #219

src/communication/RpcClient.cpp
gregmedd: Fixed

include/up-cpp/communication/RpcClient.h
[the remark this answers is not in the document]
gregmedd: Changed

src/communication/RpcClient.cpp
gregmedd: Fixed with next update to PR

gregmedd:
Note: intentionally pushing test for bug discovered in review before pushing bug fix to verify test fails in CI. See [test output](https://github.com/eclipse-uprotocol/up-cpp/actions/runs/9900262416/job/27350758263?pr=202)

src/communication/RpcClient.cpp
gregmedd: Fixed

test/coverage/communication/RpcClientTest.cpp
gregmedd: Issue #220

src/communication/RpcClient.cpp
gregmedd: Added a comment. It makes more sense now that it has been re-written, too.

src/communication/RpcClient.cpp
gregmedd: I think what I'm going to have to do to address this issue is

1. Add the request ID to the `PendingRequest` struct
2. Add a second `scrub()` method to `ScrubablePendingQueue` that takes a UUID to scrub a single request
3. Add a second `scrub()` method to `ExpireWorker` that takes a UUID and puts it in a scrub queue for deferred scrubbing (and then wakes the worker thread)
4. Modify the callback wrapper such that it calls that `scrub()` method on the `ExpireService` for the request ID inside of the `call_once` section and after the message has been passed to the caller.

Testing this is also going to be somewhat challenging. There isn't a direct way to observe if the request is still in the pending queue. I think I would need to add a protected `bool pending(v1::Uuid)` method to `RpcClient` (along with the supporting methods in the expiration service infrastructure). Then the test could inherit from RpcClient and use that `pending()` method to check.

I'm a little nervous about the scope of that change and the possible interactions this close to a release. @billpittman - how bad do you think this issue is, and could we possibly take it as an improvement in alpha.3?

gregmedd: I've preemptively filed #221 to address this as a follow-up in alpha.3. Let me know what you think - can we cut the alpha.2 release with this issue or is it a critical bug?

billpittman: I agree this could have several ripples and isn't really worth holding the release.  Assuming the issue you filed doesn't get downgraded to the point of obscurity, I'm fine with waiting until the next open window to fix this.

billpittman: I'd say this conversation is resolved.

gregmedd: I think you'll find I don't like letting issues like this fade away :grin:

billpittman (approved):
Thanks for the changes.

PR #203, opened 2024-07-09
Publish Subscribe integrated test

ruchirchauhan:
This test tests the complete pub-sub flow to make sure data flows correctly through the pipeline.

### DIFF
--- a/test/extra/PublisherSubscriberTest.cpp
+++ b/test/extra/PublisherSubscriberTest.cpp
@@ -9,29 +9,101 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include <google/protobuf/util/message_differencer.h>
 #include <gtest/gtest.h>
 
-namespace {
+#include "UTransportMock.h"
+#include "up-cpp/communication/Publisher.h"
+#include "up-cpp/communication/Subscriber.h"
+#include "up-cpp/datamodel/serializer/UUri.h"
 
-class TestFixture : public testing::Test {
+using namespace uprotocol::communication;
+using namespace uprotocol::v1;
+using namespace uprotocol::datamodel::serializer::uri;
+using MsgDiff = google::protobuf::util::MessageDifferencer;
+
+namespace {
+using namespace uprotocol::datamodel::builder;
+
+class TestPublisherSubscriber : public testing::Test {
 protected:
 	// Run once per TEST_F.
 	// Used to set up clean environments per test.
-	void SetUp() override {}
+	void SetUp() override {
+		source_.set_authority_name("10.0.0.1");
+		source_.set_ue_id(0x00011101);
+		source_.set_ue_version_major(0xF1);
+		source_.set_resource_id(0x0);
+
+		topic_.set_authority_name("10.0.0.1");
+		topic_.set_ue_id(0x10010001);
+		topic_.set_ue_version_major(0xF8);
+		topic_.set_resource_id(0x8101);
+		transportMock_ =
+		    std::make_shared<uprotocol::test::UTransportMock>(source_);
+
+		format_ = UPayloadFormat::UPAYLOAD_FORMAT_TEXT;
+		priority_ = UPriority::UPRIORITY_CS2;
+		ttl_ = std::chrono::milliseconds(1000);
+	}
+
 	void TearDown() override {}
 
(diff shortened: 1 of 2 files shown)


test/coverage/communication/PublisherTest.cpp
gregmedd, line 1: These tests should be in `test/extra/PublisherSubscriberTest.cpp`

test/coverage/communication/PublisherTest.cpp
```diff
@@ -162,4 +165,36 @@ TEST_F(TestPublisher, PublishSuccessWithoutPriority) {
...
 }
 
+TEST_F(TestPublisher, PubSubSuccess) {
+	// sub
+	auto transportSub =
+	    std::make_shared<uprotocol::test::UTransportMock>(source_);
+
+	auto callback = [this](auto arg1) {
+		return this->handleCallbackMessage(arg1);
+	};
```
gregmedd, line 175: Minor: the message could just be captured directly here instead of adding a member to the test class

```suggestion
    uprotocol::v1::UMessage captured_message
	auto callback = [&captured_message](auto message) {
		captured_message = message;
	};
```

test/coverage/communication/PublisherTest.cpp
```diff
@@ -162,4 +165,36 @@ TEST_F(TestPublisher, PublishSuccessWithoutPriority) {
...
+
+	uprotocol::v1::UStatus retval;
+	retval.set_code(uprotocol::v1::UCode::OK);
+	transportMock_->send_status_ = retval;
+
+	Payload testPayload(testPayloadStr, format_);
+	auto status = publisher.publish(std::move(testPayload));
+
+	// // Manually bridge the two transports
+	transportSub->mockMessage(transportMock_->message_);
```
gregmedd, line 193: Add a step before this confirming the source/sink of the published message match the source/sink filters of the publisher's listener registration.

gregmedd (changes requested):

test/coverage/communication/PublisherTest.cpp
ruchirchauhan: Done

test/coverage/communication/PublisherTest.cpp
ruchirchauhan: Done

test/coverage/communication/PublisherTest.cpp
ruchirchauhan: done

test/coverage/communication/PublisherTest.cpp
gregmedd, line 1: There should be no changes in this file

test/extra/PublisherSubscriberTest.cpp
```diff
@@ -9,29 +9,101 @@
...
+	uprotocol::v1::UStatus retval;
+	retval.set_code(uprotocol::v1::UCode::OK);
+	transportMock_->send_status_ = retval;
+
+	Payload testPayload(testPayloadStr, format_);
+	auto status = publisher.publish(std::move(testPayload));
+
+	// Test
+	EXPECT_EQ(AsString::serialize(topic_),
+	          AsString::serialize(transportSub->sink_filter_));
```
gregmedd, line 99: The URI check should probably be against the message and not the local variable:

```suggestion
	EXPECT_EQ(AsString::serialize(transportMock_->message_.sink()),
	          AsString::serialize(transportSub->sink_filter_));
```

gregmedd (commented):
One minor comment left

test/extra/PublisherSubscriberTest.cpp
ruchirchauhan: For publish message sink is empty, I have changed it to transportMock_->message_.attributes().source()

gregmedd: If that's the case, shouldn't it be checked against the source filter?

ruchirchauhan:
This should be ready to be merged

test/extra/PublisherSubscriberTest.cpp
gregmedd: @ruchirchauhan - this is still open. If publish methods can only have sources, then shouldn't they be checked against the source filter and not the sink filter?

PR #208, opened 2024-07-11
Adding linter script for valgrind

gregmedd:
Useful for developers wanting to run valgrind checks similar to those used in the CI system against their tests locally. The script will ignore test case errors and report only valgrind findings.

### DIFF
--- a/lint/valgrind.sh
+++ b/lint/valgrind.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+test_binary=$1
+
+if [ ! -x "$test_binary" ]; then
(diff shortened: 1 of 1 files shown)


billpittman (commented):
Overall, I like the change, but I do have a question about valgrind versions.  Should the script check to see that the version installed on a local dev's computer is the same version used on the CI (and flag/stop if not)?  Otherwise the usability of the script goes down.

billpittman (approved):
Still approving the change, because overall the change is good and whether we use a default file or not shouldn't block the merge.

PR #210, opened 2024-07-11
Add clang-tidy linter script

gregmedd:
The CI job, it turns out, has not actually been running clang-tidy against our code. As a result, linter issues have been sneaking in to the codebase. This script can be run to scan the entire source tree with clang-tidy.

### DIFF
--- a/lint/clang-tidy.sh
+++ b/lint/clang-tidy.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+PROJECT_ROOT="$(realpath "$(dirname "$0")/../")"
+
+if [ -n "$(which clang-tidy-12)" ]; then
+    # NOTE: Using clang-tidy-12 in CI system, too
+    LINTER=clang-tidy-12
+elif [ -n "$(which clang-tidy)" ]; then
+    echo "Did not find clang-tidy-12. Trying clang-tidy. Results may not"
+    echo "match formatting in GitHub CI process."
+    LINTER=clang-tidy
+else
+    echo "Could not find clang-tidy. Please make sure it is installed" 1>&2
+    exit 2
+fi
+
+compile_database="$1"
+
+if [ -z "$compile_database" ]; then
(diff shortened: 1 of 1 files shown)


billpittman (approved):

lint/clang-tidy.sh
```diff
@@ -0,0 +1,81 @@
...
+	usage
+	exit 1
+elif [ ! -f "$compile_database" ]; then
+	echo "Compile database file not found. Make sure cmake was configured" 1>&2
+	echo "with '-DCMAKE_EXPORT_COMPILE_COMMANDS=yes' and re-run the build" 1>&2
+	echo 1>&2
+	echo "Usage:" 1>&2
+	usage
+	exit 1
+fi
```
billpittman, line 50: I think most people compile with the same set of conan/cmake commands.  Should we make the script look for the build/(Release|Debug)/compile_commands.json files before dying and saying the file isn't found ?

gregmedd: If both `build/Release` and `build/Debug` exist, which one should it choose?

billpittman: Favor Release always, and make sure to output from the script something like:

Using compile script found at : build/Release/compile_commands.json
<clang-tidy output>

billpittman: FWIW, I just ran the up-cpp debug build instructions, and there wasn't a compile_commands.json file produced on either build.  I swear I had one before, but unfortunately I did a "rm -fr build" to start with a clean build folder.  Apparently too clean.

gregmedd: The `-DCMAKE_EXPORT_COMPILE_COMMANDS=yes` isn't included with any build by default

gregmedd: I think I'm going to hold off on making this change right now. My concern is that it might be too easy to end up in a scenario where the output of the script is misleading because it is using the wrong compile commands file. We can always change its behavior in a follow-up PR.

PR #211, opened 2024-07-11
Enable build warnings and fix the warnings

ruchirchauhan:


CMakeLists.txt
```diff
@@ -58,17 +58,26 @@ target_include_directories(${PROJECT_NAME}
...
-target_link_libraries(${PROJECT_NAME} 
-	PRIVATE 
+target_compile_options(${PROJECT_NAME} PUBLIC
+	-Wall
+	-Wcast-qual
+	-Wuninitialized
+	-Wconversion
+	-pedantic
+	-Werror
+)
```
gregmedd, line 75: Add [`-Wswitch-enum`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wswitch-enum). This will issue a warning for any switch statement on an enumerated type where one of the enumeration values is missing, _even if there is a `default:` case_.

```suggestion
target_compile_options(${PROJECT_NAME} PUBLIC
	-Wall
	-Wcast-qual
	-Wuninitialized
	-Wconversion
	-Wswitch-enum
	-pedantic
	-Werror
)
```

src/transport/UTransport.cpp
```diff
@@ -76,6 +76,8 @@ UTransport::registerListener(const v1::UUri& sink_filter,
 
 const v1::UUri& UTransport::getDefaultSource() const { return defaultSource_; }
 
-void UTransport::cleanupListener(CallableConn listener) {}
+void UTransport::cleanupListener(CallableConn listener) {
+	static_cast<void>(listener);
+}
```
gregmedd, line 81: Minor: Generally I prefer leaving the parameter unnamed to indicate it is unused:
```suggestion
void UTransport::cleanupListener(CallableConn) {}
```

test/coverage/communication/NotificationSinkTest.cpp
```diff
@@ -126,7 +126,7 @@ TEST_F(NotificationSinkTest, NotificationSinkSuccessWithoutSourceFilter) {
 	EXPECT_TRUE(MsgDiff::Equals(testTopicUUri_, transport->sink_filter_));
 
 	const size_t max_count = 100;
-	for (auto i = 0; i < max_count; i++) {
+	for (size_t i = 0; i < max_count; i++) {
```
gregmedd, line 129: Minor: an alternate way of writing this
```suggestion
	for (auto remaining = max_count; remaining > 0; --remaining) {
```
This has two benefits:

1. Replaces the `i` index name with a clear name (note: it could also be `remaining_notifications` for greater clarity)
2. `remaining` will always be the same type as `max_count`

test/coverage/communication/SubscriberTest.cpp
```diff
@@ -122,7 +122,7 @@ TEST_F(SubscriberTest, SubscribeSuccess) {
 	EXPECT_TRUE(MsgDiff::Equals(testTopicUUri_, transport->sink_filter_));
 
 	const size_t max_count = 100;
```
gregmedd, line 124: Minor: use constexpr
```suggestion
	constexpr size_t max_count = 100;
```

gregmedd (changes requested):
A few comments. The only one required for merging this PR is the one adding `-Wswitch-enum`. The rest are nice to have, but not required.

test/coverage/communication/NotificationSinkTest.cpp
ruchirchauhan: This is the warning it throws if I use auto

/home/kzpngr/Work/SDV2/opensrc/rc-forks/cpp/up-cpp/test/coverage/communication/NotificationSinkTest.cpp:129:44: error: comparison of integer expressions of different signedness: ‘int’ and ‘const size_t’ {aka ‘const long unsigned int’} [-Werror=sign-compare]
  129 |         for (auto remaining = 0; remaining < max_count; remaining++) {
      |                                  ~~~~~~~~~~^~~~~~~~~~~
cc1plus: all warnings being treated as errors

I think it should be ok to use size_t

gregmedd: The warning is correct: you can't use `auto i = 0`. The suggestion was to use `auto i = max_count` which should be fine since `max_count` has an explicit type.

gregmedd (approved):
The required change has been made. Approving.

Issue #217, opened 2024-07-11
Remove utils::IpAddress


gregmedd:

gregmedd:
Not used for the release nor is it likely to be needed. Remove

Issue #216, opened 2024-07-11
Remove utils::ThreadPool


gregmedd:

gregmedd:
Not implemented or used in 1.0.0. Remove for the release.

Issue #215, opened 2024-07-11
Remove utils::CyclicQueue


gregmedd:

gregmedd:
It is currently unused for the 1.0.0 release. This module can be removed. It can be retrieved from the history if we need it.

PR #222, opened 2024-07-12
Removing unused utility stubs

### LINKED ISSUE #214: Remove utils::Base64
It is unclear what this is needed for, and it certainly won't block the critical path.

Alternatively, it can be removed if we don't need it.

gregmedd:
These utilities were not needed for the 1.0.0 release of up-cpp, so they're being removed from the release branch.

Closes #214 #215 #216 #217


billpittman (approved):
LGTM

PR #240, opened 2024-07-17
Address spec alignment issue by revising UTransport and NotificationSink APIs

gregmedd:
The core changes here:

- Updated `UTransport` API to correct alignment on optional/required URIs when registering listeners
- Updated `NotificationSink` API to align with `UTransport` API update
- Updated listener filter URI validation to allow wildcards (via addition of `validator::uri::isValidFilter()`)
- Update validation checks for URIs (in particular, source URIs) to use the correct validator for the specific message type (deprecating `validator::uri::isValid()` in the process)
- Fixing source / sink swap in `Subscriber` and `NotificationSink`

Each change above, plus all of the unit test updates, ripple outward from the correction to the source and sink URIs in `UTransport`, `NotificationSink`, and `Subscriber`.

digital pond ripple


gregmedd:
Remaining work:

- Unit testing new APIs
- Checking if there are any unnecessary changes that caught a ride in these commits

test/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ file(COPY
 # Invoked as add_coverage_test("SomeName" sources...)
 function(add_coverage_test Name)
     add_executable(${Name} ${ARGN})
-    target_compile_options(${Name} PRIVATE -g -Og)
+    target_compile_options(${Name} PRIVATE -g -Og -Wno-deprecated-declarations)
```
gregmedd, line 28: This _could_ be `-Wno-error=deprecated-declarations` instead to preserve the warnings when the test code is built.

src/transport/UTransport.cpp
```diff
@@ -41,40 +40,82 @@ v1::UStatus UTransport::send(const v1::UMessage& message) {
...
 }
 
-const v1::UUri& UTransport::getDefaultSource() const { return defaultSource_; }
+// NOTE: deprecated
+utils::Expected<UTransport::ListenHandle, v1::UStatus>
+UTransport::registerListener(const v1::UUri& sink_filter,
+                             ListenCallback&& listener,
+                             std::optional<v1::UUri>&& source_filter) {
+	if (!source_filter) {
+		// Workaround for deprecated API not fully reflecting the use cases
```
stevenhartley, line 102: I don't get this at all @gregmedd please IM me so we can walk through this please.

stevenhartley (changes requested):
please see comments

include/up-cpp/datamodel/validator/UUri.h
```diff
@@ -73,12 +89,17 @@ using ValidationResult = std::tuple<bool, std::optional<Reason>>;
...
-///       improved readability where it is called.
-[[nodiscard]] ValidationResult isValidDefaultSource(const v1::UUri&);
+/// @note The requirements for this URI are identical to isValidRpcResponse()
+///       except that the authority name is not allowed to be blank.
+[[nodiscard]] ValidationResult isValidDefaultEntity(const v1::UUri&);
+
+/// @note DEPRECATED
+/// @see isValidDefaultEntity()
+/// @brief Checks if UUri is valid as a default source on a UTransport.
+[[deprecated]] [[nodiscard]] ValidationResult isValidDefaultSource(
```
billpittman, line 101: In the deprecated isValidDefaultSource, can we also get "deprecated" to add a string to say "Use isValidDefaultEntity instead" ??

include/up-cpp/datamodel/validator/UUri.h
```diff
@@ -49,7 +57,11 @@ std::string_view message(Reason);
 ///     }
 using ValidationResult = std::tuple<bool, std::optional<Reason>>;
 
-/// @brief Checks if UUri is a valid UUri of any format.
+/// @brief Checks if UUri is a valid UUri for use as an attribute in a message
+///
+/// ... but maybe this shouldn't exist. What is allowable is determined by the
+/// message type and is checked using the message validator. I don't think
+/// there is actually a usecase for this anymore.
```
billpittman, line 64: Agreed, I like the idea of the message validator checking the UUri's function/type/use explicitly.  Removing this interface seems like a good choice.  I wish there was something between deprecated and removed, like a very-deprecated.

src/communication/RpcServer.cpp
```diff
@@ -108,7 +106,18 @@ v1::UStatus RpcServer::connect(const v1::UUri& method, RpcCallback&& callback) {
 			    // Ignoring status code for transport send
 			    std::ignore = transport_->send(response);
 		    }
-	    });
+	    },
+	    // source_filter=
+	    []() {
+		    v1::UUri any_uri;
+		    any_uri.set_authority_name("*");
+		    any_uri.set_ue_id(0xFFFFFFFF);
```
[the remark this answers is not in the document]
gregmedd: You are correct :sweat_smile: 

Thanks for catching that!

src/transport/UTransport.cpp
gregmedd: @stevenhartley - I sent you a message, but we didn't manage to catch time to talk. We did go over this and discuss it amongst the developers and agreed that, while a bit ugly, this is probably the right way to handle the deprecated API.

The summary is that we are doing two things to fix the bug:

1. Deprecating the old API so it can be replaced with one that is better aligned with the spec
2. Modifying the behavior of the deprecated API so that it fixes the underlying bug in code that is currently using it.

We do not plan on keeping deprecated APIs forever, but in the short term this helps minimize the impact of changes we are making after alpha.2

include/up-cpp/datamodel/validator/UUri.h
gregmedd: Since many uses of up-cpp are likely to have `-Werror` enabled, it's likely that this will have the effect of being "very deprecated" - the compiler will produce an error and refuse to compile calls to deprecated methods.

On a side note, I should revise that comment.

gregmedd:
@stevenhartley @billpittman - Changes have been made to this PR. Please confirm your comments have been addressed and approve. Thanks!

billpittman (approved):
LGTM - Thanks for addressing my comments.

stevenhartley (approved):
LGTM

PR #256, opened 2024-08-01
Add clang-tidy-12 in CI

BishrutSubedi:
This PR adds clang-tidy-12 in CI pipeline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -416,14 +416,14 @@ jobs:
           sudo apt-get install -y clang-tidy-12
 
       - name: Run clang-tidy on source
-        working-directory: up-cpp
+        working-directory: .
         run: |
           find src include -name '*.cpp' -or -name '*.hpp' | \
           xargs -I {} clang-tidy -p build/Release/compile_commands.json {} \
           2>&1 | tee clang_tidy_source_report.txt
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -410,34 +410,44 @@ jobs:
         with:
           path: up-cpp
 
-      - name: Run linters on source
-        id: source-linter
-        uses: cpp-linter/cpp-linter-action@v2
```
gregmedd, line 415: This tool runs both clang-format and clang-tidy, and adds annotations to CI jobs / PR files. If possible, we should continue to use it for at least clang-format (and ideally clang-tidy too)

BishrutSubedi: Sounds good, I'll investigate the existing tool, vs installing clang-tidy-12 in CI.

The cpp-linter-action is will integrated with Github for adding annotations to CI jobs. It works well with clang-format, and we should continue using it. For clang-tidy, we might want to consider running clang-tidy separately if we need more detailed control or output. I will investigate both this tool, if this setup meets our goal then continuing using this tool would be much efficient and simpler.

gregmedd:
@BishrutSubedi - can this be closed now that #258 has been merged?

BishrutSubedi:
Yes, good to close

PR #261, opened 2024-08-06
Shared_ptr<UTransport> nullptr check

ruchirchauhan:
All code that takes a shared_ptr<UTransport> parameter, now check for nullptr.

### DIFF
--- a/src/communication/NotificationSource.cpp
+++ b/src/communication/NotificationSource.cpp
@@ -23,6 +23,10 @@ NotificationSource::NotificationSource(
       notify_builder_(
           UMessageBuilder::notification(std::move(source), std::move(sink))
               .withPriority(priority.value_or(v1::UPriority::UPRIORITY_CS1))) {
+	if (!transport_) {
+		throw transport::NullTransport("transport cannot be null");
+	}
+
 	if (payload_format.has_value()) {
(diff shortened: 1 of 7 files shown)


src/communication/NotificationSource.cpp
```diff
@@ -35,11 +39,19 @@ NotificationSource::NotificationSource(
 v1::UStatus NotificationSource::notify(
     datamodel::builder::Payload&& payload) const {
 	auto message = notify_builder_.build(std::move(payload));
+	if (!transport_) {
```
gregmedd, line 42: Is this check necessary here? Is it possible to have a null transport_ member if it's checked before construction?

gregmedd:
I'm wondering if this shouldn't target the 1.0 branch so it can be included in the 1.0.1 bugfix release

gregmedd:
@ruchirchauhan - please rebase this PR to the `v1.0_up-v1.6.0` branch. I have already updated the merge target for the PR

src/communication/NotificationSource.cpp
ruchirchauhan: It's not I agree. Removed

gregmedd (approved):

PR #275, opened 2024-08-09
RPC client/server test using UTransportMock

debruce:
This test demonstrates up-cpp RPC code in a round trip test using UTransportMock. This test is not working with a single transport handle. It will work with two transport handles sharing the same identity. It not clear if this is a problem with UTransportMock or an unexpected issue in up-cpp code.

### DIFF
--- a/test/extra/RpcClientServerTest.cpp
+++ b/test/extra/RpcClientServerTest.cpp
@@ -9,21 +9,62 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include <UTransportMock.h>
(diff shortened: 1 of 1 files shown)


gregmedd (approved):
Looks good to me. @debruce - Would you be able to rebase this branch onto the `v1.0_up-v1.6.0` branch? I would like to include this test with the bugfix release.

debruce:
I think this is correctly rebased to v1.0-up-v.1.6.0, and ready to merge or review.

gregmedd (approved):

PR #276, opened 2024-08-09
Bringing 1.0 bugfixes back to main

### LINKED ISSUE #193: All code that takes a `shared_ptr<UTransport>` parameter should check for `nullptr`
This mostly applies to layer 2 communication modules. At the interfaces where the shared pointer is received, it should be checked for `nullptr`. Not all modules do this.

We may want to add a `NullTransport` exception derived from `std::invalid_argument`.

Tests should be added to cover this case.

gregmedd:
Originally merged to 1.0 branch in #261

closes #193


gregmedd:
Linting results on main are tracked in a separate set of issues. This PR can proceed without resolving these findings.

PLeVasseur (approved):
Approving since these are all bug fixes being brought back to main

PR #290, opened 2024-08-14
Merging RPC Client / Server test to main

### LINKED ISSUE #174: Write extra RpcClientTest for up-cpp
The “extra” tests provide for more integrated testing than lower-level “coverage” unit tests. They check the interoperation between components. However, they do not count towards code coverage as their assertions are made at a higher level where detailed behaviors may be missed.

This test will exercise the interoperability between RpcClient and RpcServer. Are messages produced by one compatible with the other? Does the servers’s listener registration match up with the URIs in a request (i.e. does everything end up in the expected spots)? And in the reverse direction for responses? Do responses find their way to the right place? Does everything validate successfully after being passed through all stages?

gregmedd:
Closes #174 

Previously reviewed and approved in #275


PLeVasseur (approved):
:+1:

PR #292, opened 2024-08-15
Adding consumer for uSubscription service

agosh01:
This PR is to implement consumer api which act as L3 interface for the user to do subscribe request to uSubscription service 

Note: Current uSubscription proto file is not usable by C++ as its missing build configuration.  In order for this PR to work up-core-api needs to updated. [See](https://github.com/eclipse-uprotocol/up-spec/pull/235)

### DIFF
--- a/include/up-cpp/client/usubscription/v3/Consumer.h
+++ b/include/up-cpp/client/usubscription/v3/Consumer.h
@@ -1,0 +1,217 @@
+// SPDX-FileCopyrightText: 2024 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache License Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef UP_CPP_CLIENT_Consumer_H
+#define UP_CPP_CLIENT_Consumer_H
+
+#include <up-cpp/communication/NotificationSink.h>
+#include <up-cpp/communication/RpcClient.h>
+#include <up-cpp/communication/Subscriber.h>
+#include <up-cpp/datamodel/builder/Payload.h>
+#include <up-cpp/utils/ProtoConverter.h>
+#include <uprotocol/core/usubscription/v3/usubscription.pb.h>
+#include <uprotocol/v1/umessage.pb.h>
+
+#include <utility>
+
+namespace uprotocol::client::usubscription::v3 {
+using namespace uprotocol::core::usubscription::v3;
+using namespace uprotocol::utils;
+
+/**
+ * @struct ConsumerOptions
+ * @brief Additional details for uSubscription service.
+ *
+ * Each member represents an optional parameter for the uSubscription service.
+ */
+struct ConsumerOptions {
+	/// Permission level of the subscription request
+	std::optional<uint32_t> permission_level;
+	/// TAP token for access.
+	std::optional<std::string> token;
+	/// Expiration time of the subscription.
+	std::optional<std::chrono::system_clock::time_point> when_expire;
+	/// Sample period for the subscription messages in milliseconds.
+	std::optional<std::chrono::milliseconds> sample_period_ms;
+	/// Details of the subscriber.
+	std::optional<google::protobuf::Any> subscriber_details;
+	/// Details of the subscription.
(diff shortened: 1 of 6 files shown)


include/up-cpp/client/usubscription/SubscriptionClient.h
```diff
@@ -0,0 +1,91 @@
...
+	/// @param callback Function that is called when a message is received.
+	/// @param ttl Time to live for the subscription.
+	/// @param sample_period_ms Sample period of messages for remote only
+	/// subscriptions.
+	/// @param subscription_details Additional details about the
+	/// subscription.
+	/// @param subscriber_details Additional details about the subscriber.
+	[[nodiscard]] SubscriptionClientOrStatus create(
+	    const v1::UUri& topic, ListenCallback&& callback,
+	    std::optional<std::chrono::milliseconds> ttl = {},
```
gregmedd: We need to include the permission level and token to cover everything that is in Java's call options structure. These fields will need to be passed to the RpcClient used to invoke the Subscribe RPC

```suggestion
	    std::chrono::milliseconds subscribe_request_ttl,
	    std::optional<uint32_t> permission_level = {},
	    std::optional<std::string> token = {},
```

include/up-cpp/client/usubscription/SubscriptionClient.h
gregmedd, line 1: Minor: I would move this to `include/up-cpp/client/usubscription/v3/SubscriptionClient.h`. I'm also tempted to replace "SubscriptionClient" with "Subscriber" or "Consumer" (the publish side would be "Publisher" or "Producer")

include/up-cpp/client/usubscription/SubscriptionClient.h
```diff
@@ -0,0 +1,92 @@
...
+#define UP_CPP_CLIENT_SubscriptionClient_H
+
+#include <up-cpp/communication/Subscriber.h>
+#include <uprotocol/v1/umessage.pb.h>
+
+#include <optional>
+#include <uprotocol/core/usubscription/v3/usubscription.proto>
+#include <utility>
+
+namespace uprotocol::client::usubscription {
```
gregmedd, line 22: minor:
```suggestion
namespace uprotocol::client::usubscription::v3 {
```

include/up-cpp/client/usubscription/SubscriptionClient.h
```diff
@@ -0,0 +1,92 @@
...
+/// @brief Interface for uEntities to create subscriptions.
+///
+/// Like all L3 client APIs, the SubscriptionClient is a wrapper on top of the
+/// L2 Communication APIs and USubscription service.
+struct SubscriptionClient {
```
gregmedd, line 27: I'm tempted to call this `Subscriber` or `Consumer` since `client` is already in the namespace. Not 100% sure there yet. What do you think?

include/up-cpp/client/usubscription/SubscriptionClient.h
```diff
@@ -0,0 +1,92 @@
...
+	    std::optional<uint32_t> sample_period_ms = {},
+	    std::optional<google::protobuf::Any>&& subscription_details = {},
+	    std::optional<google::protobuf::Any>&& subscriber_details = {});
+
+	/// @brief get subscription response
+	///
+	/// @returns SubscriptionResponse provided by uSubscription service
+	SubscriptionResponse getSubscriptionResponse() const {
+		return subscription_response_;
+	}
```
gregmedd, line 57: I'm not sure if this is _really_ needed - we can probably handle the response internally and provide just a status outward. I'll need to think on it a little more.

include/up-cpp/client/usubscription/SubscriptionClient.h
agosh01: Subscriber may cause confusion between L2 Subscriber just in context of writing and talking. 
Might be better to call it Consumer unless Consumer might be used elsewhere too

include/up-cpp/client/usubscription/SubscriptionClient.h
agosh01: `message SubscriptionResponse {
  // Current status of the subscription
  SubscriptionStatus status = 1;

  // Any platform (uDevice) specific additional event delivered configuration
  // information for how the subscriber should consume the published events.
  EventDeliveryConfig config = 2;

  // Subscription topic passed to SubscriptionRequest
  uprotocol.v1.UUri topic = 3;
}`

As SubscriptionResponse may contain configuration information  which might be critically to consume information

PLeVasseur:
Handled by #305

Issue #270, opened 2024-08-07
clang-tidy: UuidBuilderTest - signedness changes


billpittman:
Just showing the error output:

/data/ultifi/up-cpp/test/coverage/datamodel/UuidBuilderTest.cpp:73:12: error: implicit conversion changes signedness: 'long' to 'unsigned long' [clang-diagnostic-sign-conversion]
/data/ultifi/up-cpp/test/coverage/datamodel/UuidBuilderTest.cpp:75:12: error: implicit conversion changes signedness: 'long' to 'unsigned long' [clang-diagnostic-sign-conversion]

gregmedd:
#271 can probably be resolved at the same time

PLeVasseur:
Closed via #316

PR #295, opened 2024-08-15
UuidBuilderTest.cpp: Use excplicit type not auto

billpittman:
Changed type to an explicit type, instead of using auto to remove a clang-tidy error.

Closes issue #270


PLeVasseur:
Previous attempts at cleaning up clang-tidy, addressed in other PRs. Closing

PR #299, opened 2024-08-15
Merge release branch to main

gregmedd:
This merge is primarily so that merges from the release branch to main in the future won't have conflicts, while also adjusting for the more updated releases used on main.


PLeVasseur (approved):
Seems a good idea 🙂

PR #301, opened 2024-08-17
Adding exception type for empty callback functions

### LINKED ISSUE #194: The callback connection module should check `std::function` parameters from outside for validity
We should not allow a connection to be established with null function objects. [See reference](https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool).

It may be worth adding a `NullFunction` exception derived from `std::invalid_argument`.

gregmedd:
This exception will be thrown when an empty std::function is passed to the callback connection module. Adding the type first so that tests can be written.

closes #194


agosh01 (approved):
LGTM

PLeVasseur:
Closing as duplicate of the now-merged #304

Issue #300, opened 2024-08-16
Add clang-13 to the CI build


gregmedd:
Clang has different checks and can be more pedantic than gcc. We should be including clang-13 in our CI process to make sure we actually support the versions we claim in our documentation.

gregmedd:
Just as a quick check, I tried compiling up-cpp with `clang++-18` and got several errors:

```
/home/user/.conan2/p/b/up-cp85320fffd98f5/b/src/datamodel/serializer/Uuid.cpp:109:20: error: implicit conversion changes signedness: 'int' to 'uint64_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
  108 |         uuid.set_msb((unix_ts_ms << UUID_TIMESTAMP_SHIFT) |
      |                                                           ~
  109 |                      (ver << UUID_VERSION_SHIFT) | rand_a);
      |                       ~~~~^~~~~~~~~~~~~~~~~~~~~
1 error generated.
gmake[2]: *** [CMakeFiles/up-cpp.dir/build.make:216: CMakeFiles/up-cpp.dir/src/datamodel/serializer/Uuid.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
/home/user/.conan2/p/b/up-cp85320fffd98f5/b/src/communication/RpcClient.cpp:135:18: error: captured structured bindings are a C++20 extension [-Werror,-Wc++20-extensions]
  135 |         auto wrapper = [callable, reqid = std::move(reqid),
      |                         ^
/home/user/.conan2/p/b/up-cp85320fffd98f5/b/src/communication/RpcClient.cpp:127:25: note: 'callable' declared here
  127 |         auto [callback_handle, callable] =
      |                                ^
/home/user/.conan2/p/b/up-cp85320fffd98f5/b/src/communication/RpcClient.cpp:140:38: error: captured structured bindings are a C++20 extension [-Werror,-Wc++20-extensions]
  140 |                                 std::call_once(*callback_once, [&callable, &m]() {
      |                                                                  ^
/home/user/.conan2/p/b/up-cp85320fffd98f5/b/src/communication/RpcClient.cpp:127:25: note: 'callable' declared here
  127 |         auto [callback_handle, callable] =
      |                                ^
/home/user/.conan2/p/b/up-cp85320fffd98f5/b/src/communication/RpcClient.cpp:148:38: error: captured structured bindings are a C++20 extension [-Werror,-Wc++20-extensions]
[...]

Issue #298, opened 2024-08-15
Investigate clang-tidy false positives


gregmedd:

debruce:
I would like to work on this.

lammjo:
PR #310 fixes the warning and errors in source files, but I suggest tackling the test files in a separate issue, #311. Afterwards, it should be possible to include building with clang-13 as well, proposed in #300.

PLeVasseur:
Sounds reasonable to me @lammjo.

PLeVasseur:
I opened #312 to be a subissue of #298 for what's tackled in #310. Also made #311 a subissue of #298.

PLeVasseur:
This was the umbrella issue and can thus also be closed!

PR #304, opened 2025-02-11
Bugfix/1.0.2/194 std function checks

lukas-he:


include/up-cpp/utils/CallbackConnection.h
```diff
@@ -254,11 +266,21 @@ struct [[nodiscard]] CalleeHandle {
...
 		}
+
 		if (!callback_) {
 			throw BadConnection(
 			    "Attempted to create a connected CalleeHandle with bad "
 			    "callback pointer");
 		}
+
+		const auto& callback_obj = *callback_;
+		if (!callback_obj) {
```
PLeVasseur, line 277: TIL!

[`std::function<R(Args...)>::operator bool`](https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool): "Checks whether *this stores a callable function target, i.e. is not empty."

test/coverage/communication/NotificationSinkTest.cpp
```diff
@@ -216,14 +216,23 @@ TEST_F(NotificationSinkTest, NullCallback) {
...
+		                             std::move(nullptr), testTopicUUri_);
+	};
+
+	using namespace uprotocol::utils;
+
+	EXPECT_THROW(test_create_nullptr(), callbacks::EmptyFunctionObject);
+
+	// Default construct a function object
+	auto test_create_empty = [transport, this]() {
+		std::ignore = NotificationSink::create(
```
PLeVasseur, line 231: TIL also about [`std::ignore`](https://en.cppreference.com/w/cpp/utility/tuple/ignore) 😅 

> Some code guides recommend using std::ignore to avoid warnings from unused return values of [[nodiscard]] functions, even though an assignment isn't required.

include/up-cpp/utils/CallbackConnection.h
```diff
@@ -234,6 +234,18 @@ struct BadConnection : public std::runtime_error {
 	    : std::runtime_error(std::forward<Args>(args)...) {}
 };
 
+/// @brief Thrown if an empty std::function parameter was received
+///
+/// A std::function can be empty. When an empty function is invoked, it will
+/// throw std::bad_function_call. We can check earlier by casting the function
+/// to a boolean. If the check fails, EmptyFunctionObject is thrown. This makes
+/// the error appear earlier without waiting for invokation to occur.
```
PLeVasseur, line 242: I wonder... it's called [`std::invoke`](https://en.cppreference.com/w/cpp/utility/functional/invoke), but on the other hand the normal English here would be invocation. I could go either way if in C++ it's standard to call this "invokation".
```suggestion
/// the error appear earlier without waiting for invocation to occur.
```

PLeVasseur (commented):
Not much to say! LGTM. I learned a bit more C++ reading through here.

I saw that a couple of stages of CI fell over. @lukas-he -- would you like to investigate?

lukas-he:
I looked into the error and as far as I understand, the linting fails before actually linting, but when trying to install clang-tidy-12 see [here](https://github.com/eclipse-uprotocol/up-cpp/actions/runs/13258786674/job/37167550589?pr=304#step:10:64). So we could go to a newer clang-tidy version. What do you think @PLeVasseur?

PLeVasseur:
> So we could go to a newer clang-tidy version.

Seems like a reasonable solution if the previous version we're using is not available and doesn't in-turn flag a bunch of more things to fix 😅 

IIRC the default runner image is now Ubuntu 24.04.1, see [here](https://github.com/eclipse-uprotocol/up-cpp/actions/runs/13258786674/job/37167550589?pr=304#step:1:4). Likely that version of clang-tidy may not be available.

If you find that updating clang-tidy flags a bunch of additional things to fix, one option is to use Ubuntu 22.04 for now. Then address the issues found by the new clang-tidy all in one PR. It's a thought!

PLeVasseur:
@lukas-he -- is the idea that we get https://github.com/eclipse-uprotocol/up-conan-recipes/pull/32 merged, then try to merge this one?

Wondering if there's anything we'd see that differs enough that up-cpp would remain broken? 🤔 

If so, I suppose I'd like to have a set of patches / PRs which update things to a working state (with proof of this with a dummy PR branch), but land them one by one.

What do you think about this approach?

lammjo:
> > So we could go to a newer clang-tidy version.
> 
> Seems like a reasonable solution if the previous version we're using is not available and doesn't in-turn flag a bunch of more things to fix 😅
> 
> IIRC the default runner image is now Ubuntu 24.04.1, see [here](https://github.com/eclipse-uprotocol/up-cpp/actions/runs/13258786674/job/37167550589?pr=304#step:1:4). Likely that version of clang-tidy may not be available.
> 
> If you find that updating clang-tidy flags a bunch of additional things to fix, one option is to use Ubuntu 22.04 for now. Then address the issues found by the new clang-tidy all in one PR. It's a thought!

I ran both clang-tidy-12 and clang-tidy-18. Interestingly, the number of warnings is actually almost the same, but the linters detect very different issues (see the occurence counts below if you're curious). I suggest we go with your suggestion of pinning the Ubuntu 22.04 image in CI and using clang-tidy-12 for now.

[clang-tidy-18-errors.txt](https://github.com/user-attachments/files/18968110/clang-tidy-18-errors.txt)
[clang-tidy-12-errors.txt](https://github.com/user-attachments/files/18968112/clang-tidy-12-errors.txt)

PLeVasseur:
> I suggest we go with your suggestion of pinning the Ubuntu 22.04 image in CI and using clang-tidy-12 for now.

Sounds good as a starting point.

PLeVasseur:
@lukas-he -- can you double-check my question over [here](https://github.com/eclipse-uprotocol/up-cpp/issues/309#issuecomment-2703823683)?

Long story short -- if #298 and #309 are one in the same, I'd say we go ahead and merge this as that had been our hand-wavey policy in the last week or so.

It heartens me to hear @lammjo mention that someone is working on #298 :)

PLeVasseur:
Merging this based on discussion with @lukas-he over Slack. Idea is to do #300, then #298 to resolve all clang-tidy issues in a couple of steps.

PLeVasseur (approved):
Thanks for the contribution @lukas-he!

PR #306, opened 2025-02-20
Payload Builder supporting google.protobuf.Any

lukas-he:
This PR reopens [PR 287](https://github.com/eclipse-uprotocol/up-cpp/pull/287). Added the code from debruce's commits manually, since the branch (and the repo) are not available. 

Add constructor to Payload. Add test case to PayloadBuilderTest to exercise this constructor by creating an Any containing a uuri, passing it to the constructor, building it, and then deserializing the Any, and the contained uuri.


PLeVasseur (approved):
A re-do of an existing PR. Merging!

PR #310, opened 2025-03-16
Fix for issue 312: fixed clang-tidy linting warnings and error in include/* and src/*

### LINKED ISSUE #312: Upgrade to clang-tidy-13, fix clang-tidy linting warnings and error in include/* and src/*

MaximilianToe:
Fixed clang-tidy warnings and errors in all files contained in the include and src directories. 
Also added the fixes suggested in Issue 264 and 270.
Additionally, updated and pinned clang-tidy version to 13 to make it consistent with the version of the clang compiler. 
Lastly, modified the clang-tidy.sh script to only check include and src. Previously, this would check all files, including cmake files, etc.
The exact files to check should probably be discussed, and the script modified accordingly. 

closes #312

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -431,6 +431,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           repo-root: up-cpp
+          version: 13
           ignore: 'test'
           style: 'file' # read .clang-format for configuration
           tidy-checks: '' # Read .clang-tidy for configuration
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -431,12 +431,11 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           repo-root: up-cpp
-          version: 13
           ignore: 'test'
           style: 'file' # read .clang-format for configuration
           tidy-checks: '' # Read .clang-tidy for configuration
           database: build/Release/compile_commands.json
-          version: 12
+          version: 13
 
           
       - name: Run linters on tests
@@ -451,7 +450,7 @@ jobs:
           style: 'file' # read .clang-format for configuration
           tidy-checks: '' # Read .clang-tidy for configuration
           database: build/Release/compile_commands.json
-          version: 12
+          version: 13
 
 
       - name: Report lint failure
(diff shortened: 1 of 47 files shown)


lammjo:
@PLeVasseur Would you be so kind to take a look at the PR and trigger the pipeline? Max is the colleague I told you about.

An open question is whether the test files should also be linted. Personally, I prefer linting tests as well. But since many of the warnings and errors in the tests are of a different nature than in the source files, I created a separate issue for these, #310, which we could tackle next.

lint/clang-tidy.sh
```diff
@@ -58,7 +58,7 @@ if [ -z "$target_source" ]; then
 	shopt -s globstar
 
 	pushd "$PROJECT_ROOT" > /dev/null
-	for f in **/*.h **/*.cpp; do
+	for f in include/**/*.h src/**/*.cpp; do # test/coverage/**/*.cpp test/extra/**/*.cpp test/include/**/*.h; do
```
lammjo, line 61: I believe there is no need to change the script in this way. If you want to limit the scope of the linter, you can provide the desired files explicitly.

.github/workflows/ci.yml
```diff
@@ -431,6 +431,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           repo-root: up-cpp
+          version: 13
```
lammjo, line 434: `version` is already defined in line 439.

MaximilianToe: Thank you for pointing this out. This should now be fixed.

PLeVasseur:
Hi @MaximilianToe -- thank you for the contribution!

I noticed that there remained 10 or so warnings / errors in the `include` and `src` folders. Is it intended these be handled in this PR?

PLeVasseur:
Hi @MaximilianToe, thanks for the contribution!

> The only remaining warnings/errors should occur in the linting job for the test files, which I would leave to another PR.

I confirmed that to be the case on the latest CI run. Great! I'll go review and approve.

include/up-cpp/utils/Expected.h
```diff
@@ -57,14 +57,18 @@ class Unexpected {
 template <typename T, typename E>
 class Expected {
 public:
-	template <typename... Args>
-	constexpr Expected(Args&&... args)
-	    : storage_(std::forward<Args>(args)...) {}
+	constexpr explicit Expected(T arg) : storage_(std::forward<T>(arg)) {}
+	// It E and T are the same type, this can cause problems. Previously, this
+	// was in use by implicid conversion
```
PLeVasseur, line 62: nit:
```suggestion
	// was in use by implicit conversion
```

PLeVasseur (approved):
Thanks again @MaximilianToe! I found one minor nit that I'll just go ahead and apply then merge.

PR #322, opened 2025-05-22
fixed instance id in RpcServer

### LINKED ISSUE #323: RpcServer listens only on default instance id
Currently the connect method of the RpcServer creats a source filter that contains only wild cards except for the resource id. Here, the default instance id 0000 is used. This inconsistent with the implementation in up-rust where the wildcard FFFF is used instead (see this method https://github.com/eclipse-uprotocol/up-rust/blob/72b6507e3b4fe0f0b3a873307a63e3d9118c130d/src/uri.rs#L377). Furthermore, this leads to problems that @lukas-he discovered when working on https://github.com/eclipse-uprotocol/up-transport-zenoh-cpp/issues/119 where this needs to be mapped to zenoh wildcards.

MaximilianToe:
This PR replaces the default instance id 0000 with the wildcard id FFFF when creating the source filter in the connect method of RpcServer. 
This fixes a bug that causes a test in the pipline of https://github.com/eclipse-uprotocol/up-transport-zenoh-cpp/pull/121 to fail. 

Closes #323

### DIFF
--- a/src/communication/RpcServer.cpp
+++ b/src/communication/RpcServer.cpp
@@ -124,9 +124,9 @@ v1::UStatus RpcServer::connect(const v1::UUri& method, RpcCallback&& callback) {
 	    []() {
 		    v1::UUri any_uri;
 		    any_uri.set_authority_name("*");
-		    // Instance ID 0 and UE ID FFFF for wildcard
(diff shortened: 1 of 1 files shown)


PLeVasseur (approved):
Thanks @MaximilianToe!

PR #324, opened 2025-06-09
Rpc client u subscription

MaximilianToe:
This pull request provides the following:
- An interface for a USubscriptionClient in USubscription.h
- An Implementation of that interface in RpcClientUSubscription.cpp/RpcClientUSubscription.h
- A template method InvokeProtoMethod to send and receive requests with a specific payload type in RpcClient.h similar to the one present in up-rust
- A helper request builder and uuri builder
- Extracting part of Consumer.cpp that was also needed for USubscription to avoid code duplication

### DIFF
--- a/include/up-cpp/datamodel/builder/UMessage.h
+++ b/include/up-cpp/datamodel/builder/UMessage.h
@@ -234,6 +234,16 @@ struct UMessageBuilder {
 	/// @return A built message with no payload populated.
 	[[nodiscard]] v1::UMessage build() const;
 
+	/// @brief Creates a UMessage based on the builder's current state.
+	///
+	/// @param A UUri of the method that should be called
+	///
+	/// @throws UnexpectedFormat if withPayloadFormat() has been previously
+	///         called.
+	///
+	/// @return A built message with no payload populated.
+	[[nodiscard]] v1::UMessage build(const v1::UUri&) const;
+
 	/// @brief Creates a UMessage with a provided payload based on the
 	///        builder's current state.
 	///
@@ -246,6 +256,21 @@ struct UMessageBuilder {
 	///
 	/// @return A built message with the provided payload data embedded.
 	[[nodiscard]] v1::UMessage build(builder::Payload&&) const;
+
+	/// @brief Creates a UMessage with a provided payload based on the
+	///        builder's current state.
+	///
+	/// @param A UUri of the method that should be called
+	///
+	/// @param A Payload builder containing a payload to embed in the message.
+	///
+	/// @note The contents of the payload builder will be moved.
+	///
+	/// @throws UnexpectedFormat if withPayloadFormat() has been previously
+	///         called and the format in the payload builder does not match.
+	///
+	/// @return A built message with the provided payload data embedded.
+	[[nodiscard]] v1::UMessage build(const v1::UUri&, builder::Payload&&) const;
 
 	/// @brief Access the attributes of the message being built.
 	/// @return A reference to the attributes of the message being built.
--- a/src/communication/RpcClient.cpp
+++ b/src/communication/RpcClient.cpp
@@ -208,19 +208,13 @@ RpcClient::InvokeHandle RpcClient::invokeMethod(v1::UMessage&& request,
 RpcClient::InvokeHandle RpcClient::invokeMethod(
     const v1::UUri& method, datamodel::builder::Payload&& payload,
     Callback&& callback) {
-	// makes a local copy of the builder to make this call thread safe
-	auto local_builder = builder_;
-	return invokeMethod(
-	    local_builder.withMethod(method).build(std::move(payload)),
-	    std::move(callback));
+	return invokeMethod(builder_.build(method, std::move(payload)),
+	                    std::move(callback));
 }
 
 RpcClient::InvokeHandle RpcClient::invokeMethod(const v1::UUri& method,
                                                 Callback&& callback) {
-	// makes a local copy of the builder to make this call thread safe
-	auto local_builder = builder_;
-	return invokeMethod(local_builder.withMethod(method).build(),
-	                    std::move(callback));
+	return invokeMethod(builder_.build(method), std::move(callback));
 }
 
 RpcClient::InvokeFuture RpcClient::invokeMethod(
--- a/src/datamodel/builder/UMessage.cpp
+++ b/src/datamodel/builder/UMessage.cpp
@@ -218,10 +218,45 @@ v1::UMessage UMessageBuilder::build() const {
 	return message;
 }
 
+v1::UMessage UMessageBuilder::build(const v1::UUri& method) const {
+	v1::UMessage message;
+	if (expectedPayloadFormat_.has_value()) {
+		throw UnexpectedFormat(
+		    "Tried to build with no payload when a payload format has been set "
+		    "using withPayloadFormat()");
+	}
+
+	*message.mutable_attributes() = attributes_;
+	*message.mutable_attributes()->mutable_sink() = method;
+	*(message.mutable_attributes()->mutable_id()) = uuidBuilder_.build();
+
+	return message;
+}
+
 v1::UMessage UMessageBuilder::build(builder::Payload&& payload) const {
 	v1::UMessage message;
 
 	*message.mutable_attributes() = attributes_;
+	*(message.mutable_attributes()->mutable_id()) = uuidBuilder_.build();
+	auto [payloadData, payloadFormat] = std::move(payload).buildMove();
+	if (expectedPayloadFormat_.has_value()) {
+		if (payloadFormat != expectedPayloadFormat_) {
+			throw UnexpectedFormat(
+			    "Payload format does not match the expected format");
+		}
+	}
+	*message.mutable_payload() = std::move(payloadData);
+	message.mutable_attributes()->set_payload_format(payloadFormat);
+
+	return message;
+}
+
+v1::UMessage UMessageBuilder::build(const v1::UUri& method,
+                                    builder::Payload&& payload) const {
+	v1::UMessage message;
+
+	*message.mutable_attributes() = attributes_;
+	*message.mutable_attributes()->mutable_sink() = method;
 	*(message.mutable_attributes()->mutable_id()) = uuidBuilder_.build();
 	auto [payloadData, payloadFormat] = std::move(payload).buildMove();
 	if (expectedPayloadFormat_.has_value()) {


include/up-cpp/client/usubscription/v3/RequestBuilder.h
```diff
@@ -0,0 +1,101 @@
...
+
+#include <utility>
+
+namespace uprotocol::core::usubscription::v3 {
+
+/// @struct USubscriptionOptions
+/// @brief Additional details for uSubscription service.
+///
+/// Each member represents an optional parameter for the uSubscription service.
+struct USubscriptionOptions {
```
PLeVasseur, line 25: Called [`CallOptions`](https://github.com/eclipse-uprotocol/up-rust/blob/b1c77dc472a31536f274a9aef6e4b1cc53ef7b2b/src/communication.rs#L118) in up-rust (sort of).

`CallOptions` includes less, actually:
```rust
pub struct CallOptions {
    ttl: u32,
    message_id: Option<UUID>,
    token: Option<String>,
    priority: Option<UPriority>,
}
```

@MaximilianToe -- could you share where you got this list of options from? Curious to know why they differ.

Should we align on `CallOptions` and then update the up-spec accordingly? Differing on naming and contents of this doesn't seem to bring value so it'd make sense to me.

cc @stevenhartley @sophokles73 for spec question visibility

include/up-cpp/utils/ProtoConverter.h
```diff
@@ -53,6 +66,125 @@ struct ProtoConverter {
...
+					v1::UStatus status;
+					status.set_code(v1::UCode::INTERNAL);
+					status.set_message(
+					    "extractFromProtobuf: Error when parsing payload from "
+					    "protobuf.");
+					return TOrStatus<T>(UnexpectedStatus(status));
+				}
+				return TOrStatus<T>(response);
+			}
+			case v1::UPayloadFormat::UPAYLOAD_FORMAT_UNSPECIFIED:
```
PLeVasseur, line 120: According to [up-spec](https://github.com/eclipse-uprotocol/up-spec/blob/main/basics/uattributes.adoc#8-payload-format), `UPAYLOAD_FORMAT_UNSPECIFIED` is a case that should most probably return an error as it's not in ProtoBuf:
image

Looking up how this is handled in up-rust, I see that it makes the same assumption about `UPAYLOAD_FORMAT_UNSPECIFIED`. Probably I'm missing something or have forgotten we made this decision 😅 

Should the up-spec be updated to note we assume it's `UPAYLOAD_FORMAT_PROTOBUF_WRAPPED_IN_ANY` if `UPAYLOAD_FORMAT_UNSPECIFIED` is given as the format? Or maybe that is written somewhere and I've missed it?

cc @sophokles73 @stevenhartley for spec question

src/client/usubscription/v3/USubscriptionUUriBuilder.cpp
```diff
@@ -0,0 +1,52 @@
...
+	    service_options.GetExtension(uprotocol::service_name);
+	const auto& service_version_major =
+	    service_options.GetExtension(uprotocol::service_version_major);
+	const auto& service_id =
+	    service_options.GetExtension(uprotocol::service_id);
+	const auto& notification_topic =
+	    service_options.GetExtension(uprotocol::notification_topic, 0);
+
+	// Set the values in the URI
+	base_uri_.set_authority_name(service_name);
```
PLeVasseur, line 33: I think the authority name must be provided, as it's not the service_name.

cc @stevenhartley @sophokles73 for spec question

src/client/usubscription/v3/USubscriptionUUriBuilder.cpp
```diff
@@ -0,0 +1,52 @@
...
+// terms of the Apache License Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "up-cpp/client/usubscription/v3/USubscriptionUUriBuilder.h"
+
+namespace uprotocol::core::usubscription::v3 {
+
+USubscriptionUUriBuilder::USubscriptionUUriBuilder() {
```
PLeVasseur, line 16: I think we'd need a means to specify the instance ID, as we will have one instance of the uSubscription Service per physical and logical device.

cc @stevenhartley @sophokles73 for spec question

PLeVasseur (commented):
Hey @MaximilianToe -- thanks for contributing!

I noted more up-spec-related items and have tagged @sophokles73 and @stevenhartley onto the PR as there may be gaps in the spec (or in my understanding of it!)

include/up-cpp/client/usubscription/v3/RpcClientUSubscription.h
```diff
@@ -0,0 +1,138 @@
...
+	///
+	/// @param transport Transport used to send messages
+	explicit RpcClientUSubscription(
+	    std::shared_ptr<transport::UTransport> transport);
+
+	~RpcClientUSubscription() override = default;
+
+private:
+	std::shared_ptr<transport::UTransport> transport_;
+	// Currently a single RpcClient can only send messages to a fixed UUri.
```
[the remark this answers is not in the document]
sophokles73: Unless there are C++ specific constraints that make this way of implementing it desirable/preferable, I would also argue in favor of up-rust's way of doing it.

The spec does not prescribe either way or the other. FMPOV it is only concerned with how each individual message exchange is expected to work.

include/up-cpp/communication/RpcClient.h
```diff
@@ -162,6 +175,51 @@ struct RpcClient {
...
+
+		auto handle = invokeMethod(
+		    builder_.build(std::move(tmp_payload)),
+		    [result_promise](const MessageOrStatus& message_or_status) {
+			    if (!message_or_status.has_value()) {
+				    result_promise->set_value(ResponseOrStatus<T>(
+				        UnexpectedStatus(message_or_status.error())));
+				    return;
+			    }
+			    auto response_or_status =
```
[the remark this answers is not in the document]
sophokles73: The response can contain arbitrary bytes in its payload. The payload_format attribute of the response message's UAttributes indicates the encoding of the payload.

include/up-cpp/client/usubscription/v3/RequestBuilder.h
sophokles73: The Communication Layer API spec had been defined rather vaguely on purpose. In order to leave _lots of legroom_ for implementors, IIRC.
I would also be fine with defining it using the same syntax/notation that we used for the Transport Layer API ...

include/up-cpp/client/usubscription/v3/RpcClientUSubscription.h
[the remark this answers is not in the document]
MaximilianToe: I agree with Kai here, the way rust implements this seems preferable to me. I was actually quite surprised when I found out during implementation that you could not reuse the rpc client because of the uuri constraint. If that constraint would be removed, I would rework this part of the code and only use a single instance of the rpc client.

 I won't be in the meeting on Thursday but @lennartbecker-d, who has also worked on this, will be there.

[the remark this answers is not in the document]
PLeVasseur: Cool, so let's consider that RpcClient will be reworked to be like up-rust to allow multiple invokable methods. If there are questions we can tackle them in the uProtocol weekly with @lennartbecker-d 👍

include/up-cpp/client/usubscription/v3/RpcClientUSubscription.h
```diff
@@ -0,0 +1,138 @@
...
+
+#include <up-cpp/communication/NotificationSink.h>
+#include <up-cpp/communication/RpcClient.h>
+#include <up-cpp/transport/UTransport.h>
+#include <uprotocol/core/usubscription/v3/usubscription.pb.h>
+
+#include "up-cpp/client/usubscription/v3/USubscription.h"
+#include "up-cpp/client/usubscription/v3/USubscriptionUUriBuilder.h"
+
+/// The uEntity (type) identifier of the uSubscription service.
```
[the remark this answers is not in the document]
lennartbecker-d: @PLeVasseur if you only refer to USUBSCRIPTION_TYPE_ID, this originates from usubscription.rs and is not used as far as I know. It was added to make it comparable.

include/up-cpp/utils/ProtoConverter.h
PLeVasseur: Cool, thanks @MaximilianToe!

src/client/usubscription/v3/USubscriptionUUriBuilder.cpp
sophokles73: That is true, but this also means, that it should be possible to simply assume the default instance ID (0x0000).

src/client/usubscription/v3/USubscriptionUUriBuilder.cpp
sophokles73: The authority name is indeed not the same thing as the service name. In fact, the service name is not part of a UUri at all. Its main purpose is to have a descriptive name available that can be used in log files or UIs.

src/client/usubscription/v3/USubscriptionUUriBuilder.cpp
PLeVasseur: That's true actually 👍 The fact that there's 1 x subscription manager per device will allow discrimination via authority, so they could always have the default instance ID.

sophokles73:
@PLeVasseur would you mind taking another look?

include/up-cpp/communication/RpcClient.h
```diff
@@ -175,25 +175,31 @@ struct RpcClient {
 	///          * A UMessage containing the response from the RPC target.
 	[[nodiscard]] InvokeFuture invokeMethod();
```
PLeVasseur, line 176: In order to have a similar API to the Rust Communication / uP-L2 APIs, this would require `invokeMethod()` to also take the method to invoke and we should align how an `RpcClient` is constructed with Rust as well to avoid the need to specify a method UUri up-front.

Does that make sense?

PLeVasseur (commented):
Thank you for the updates @MaximilianToe! Can you take a look at the comments I left?

It feels to me there's still a bit of alignment work to ensure that the C++ RpcClient API mimics the Rust equivalent. Happy to hear your thoughts.

include/up-cpp/client/usubscription/v3/RpcClientUSubscription.h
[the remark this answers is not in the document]
PLeVasseur: I think this is still open due to `invokeMethod()` so leaving as unresolved for now.

include/up-cpp/communication/RpcClient.h
```diff
@@ -175,25 +175,31 @@ struct RpcClient {
...
 
-	template <typename T, typename R>
-	InvokeProtoFuture<T> invokeProtoMethod(const R& request_message) {
-		auto result_promise =
-		    std::make_shared<std::promise<ResponseOrStatus<T>>>();
-		auto future = result_promise->get_future();
-
+	template <typename R>
+	InvokeHandle invokeProtoMethod(const R& request_message,
```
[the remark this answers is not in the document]
MaximilianToe: Thanks for bringing up all these valid questions. I will go through them top to bottom:
1. Yes, this method is exposed to users to up-cpp.
2. The idea was to have a method available to the user that works like invokeMethod(Payload, callback) but works directly with protobuf objects.
3. That is a good question. The reason why I choose this name was because it accepts a protobuf object as input (in fact, thats also the reason for the other methods names), but I see that this might be confusing. 
4. The proto in the name actually referes to the input. But I agree, there should be a difference in the name, depending on the expected kind of payload. 
5. Same as the above. 

I think you make a good point with your points raised concering the naming. After thinking about it, I think the Proto in the name should depend on the expected output, not the input. About point 1. and 2., this is how I intended it to be, but if you think this is a bad idea, I am of course open to change this.

include/up-cpp/client/usubscription/v3/RequestBuilder.h
MaximilianToe: This struct has been renamed to CallOptions to follow the same naming as is used in up-rust.

include/up-cpp/communication/RpcClient.h
PLeVasseur: > Would you suggest to do this alignment as part of this PR or rather defer to a separate issue + PR?

So depending on the context I would probably normally just suggest to:
> defer to a separate issue + PR

However, in this case, if we merge in this current state we have a commit in main which has kind of this mixed API. So -- two paths in my mind:
1. create an issue, create a separate PR which would go in ahead of this one which updates the existing methods to work with a single `RpcClient` to service multiple invokable methods, rebase your PR's branch on main
2. make the issue, add the implementation work in the PR, tag in the body of the PR that you're also including work from the new issue

Which path makes most sense to you?

MaximilianToe: I like keeping things separate, this also prevents the current PR from getting even larger. So I created a separate issue and proceed as suggestion 1. :) 
https://github.com/eclipse-uprotocol/up-cpp/issues/325

include/up-cpp/client/usubscription/v3/RpcClientUSubscription.h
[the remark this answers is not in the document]
PLeVasseur: Marking as resolved, since this behavior is now consistent. ☺️

include/up-cpp/communication/RpcClient.h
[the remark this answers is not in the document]
PLeVasseur: This was completed and renaming done, yay.

PLeVasseur:
> > Okay -- I'll ask about this on Slack 🫡
> 
> I have already [created a PR](https://github.com/eclipse-uprotocol/.eclipsefdn/pull/87) to reduce the number of required reviews to 1 again ...

Thanks! It does still seem to show as needing two reviews on this one tho 🤔 

image

Is there some lingering effect on already-existing PRs or something like that?

sophokles73:
> > > Okay -- I'll ask about this on Slack 🫡
> > 
> > 
> > I have already [created a PR](https://github.com/eclipse-uprotocol/.eclipsefdn/pull/87) to reduce the number of required reviews to 1 again ...
> 
> Thanks! It does still seem to show as needing two reviews on this one tho 🤔
> 
> image
> Is there some lingering effect on already-existing PRs or something like that?

Maybe, I don't know. Anyway, I will simply approve this PR now and then you can squash and merge it ...

sophokles73 (approved):
LGTM

PR #329, opened 2025-12-03
Minor changes for alpha4

### LINKED ISSUE #214: Remove utils::Base64

MaximilianToe:
Added some minor improvements for the alpha4-release. These include the following:
 - Deleted some unused files. This closes #214 and #216. 
 - Added missing doc strings.
 - (optional, commit could be removed if not desired) Renamed invokeMethodToProto to invokeMethodFromProto for the cases where the input of the method in a protobuf object and the return value is either a Handle or a uMessage. This would name would better reflect the behaviour.


include/up-cpp/communication/RpcClient.h
```diff
@@ -180,10 +180,24 @@ struct RpcClient {
...
+	///         received before the request expired (based on request TTL).
+	///       * A UStatus with the value returned by UTransport::send().
+	///       * A UStatus based on the commstatus received in the response
+	///         message (if not OK).
+	///       * A UMessage containing the response from the RPC target.
 	template <typename R>
-	InvokeHandle invokeMethodToProto(const v1::UUri& method,
-	                                 const R& request_message,
-	                                 Callback&& callback) {
+	[[nodiscard]] InvokeHandle invokeMethodFromProto(const v1::UUri& method,
```
PLeVasseur, line 198: Makes sense, good thought

PLeVasseur (approved):
Beautiful to see the documentation fill in! 🥰 Thank you for getting this ready.

